PR #38, opened 2016-08-16
Feature/loading improvements and bug fixes

emoralesb05:
## Description

Second iteration on TdLoadingDirective/Service/Component.

### What's included?

- Added `width` to loading circle to be in par with circle height.
- Removed deprecated methods (soon to be deprecated) and made loading component creation sync.
- Added max diameter and height for loading circle.
- Updated loading docs usage
- Used `NgZone#runOutsideAngular` to address a bug with protractor where it gets stuck waiting when adding attaching/detaching the component and doing its animations.

Known issue:
- May need to call change detection when registering loading in a component life cycle hook. https://github.com/angular/angular/issues/6005
e.g.
```
ngAfterViewInit(): void {
    this.registerLoadingReplace();
    this._changeDetectorRef.detectChanges();
  }
```

#### Test Steps

- [x] `ng serve`
- [x] `Go to loading component`

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -4,7 +4,7 @@ import { ROUTER_DIRECTIVES } from '@angular/router';
 import { MdIcon } from '@angular2-material/icon';
 import { MD_LIST_DIRECTIVES } from '@angular2-material/list';
 
-import { TdLayoutComponent } from '../platform/core';
+import { TdLayoutComponent, TD_LOADING_ENTRY_COMPONENTS } from '../platform/core';
 
 @Component({
   directives: [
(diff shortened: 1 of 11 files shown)


src/app/app.component.ts
```diff
@@ -17,6 +17,7 @@ import { TdLayoutComponent } from '../platform/core';
   selector: 'td-docs-app',
   styleUrls: ['app.component.css'],
   templateUrl: 'app.component.html',
+  precompile: [ TD_LOADING_ENTRY_COMPONENTS ],
```
aciccarello: Haven't used `precompile` before but it looks like it is [being renamed](https://github.com/angular/angular/commit/6f4e49e). The change is in angular master ([Component metadata docs](https://angular.io/docs/ts/latest/api/core/index/ComponentMetadata-class.html)) and I presume RC5 so #37 will have to be updated.

emoralesb05: Yes, its actually just for rc4, in rc5 it uses entryComponents in ngModule. I already have that already in https://github.com/Teradata/covalent/pull/37, but we are waiting on that since rc5 and CLI have some stability issues in this release.

PR #44, opened 2016-08-18
Feature/covalent chips

emoralesb05:
## Description

First iteration of the Chips Module.

#### TdChipsComponent

Added <td-chips> component that generates a list of string as chips.
- Module docs and demo.
- Added package.json
- Added README.md
- [items] attribute to provide a search list for autocomplete.
- [readOnly] flag to disable edition (entering/removing chips)
- [requireMatch] flag to validate autocomplete input against search list.
- Basic `a11y` (tabbing between autocomplete and removal buttons, and enter to remove chip)

Usage:
```html
<td-chips [items]="items" [(ngModel)]="model" readOnly="readOnly" (add)="add($event)" (remove)="remove($event)" requireMatch>
</td-chips>
```
Properties:
image

#### Test Steps

- [x] `ng serve`
- [x] go to http://localhost:4200/#/components/chips
- [x] Use demo and see documentation.
- [x] Try to use element following documentation anywhere in docs app.
- [x] :v: 

td-chips-demo1

### DIFF
--- a/src/app/components/components/chips/chips.component.html
+++ b/src/app/components/components/chips/chips.component.html
@@ -29,7 +29,7 @@
   <md-card-content>
     <h2><code><![CDATA[<td-chips>]]></code></h2>
     <p>Use <code><![CDATA[<td-chips>]]></code> element to generate a list of strings as chips.</p>
-    <p>Add the [items] attribute to enable the autocomplete with a search list, and [requireMatch] to validated the input against the provided search list.</p>
+    <p>Add the [items] attribute to enable the autocomplete with a search list, and [requireMatch] to validate the input against the provided search list.</p>
     <h3>Properties:</h3>
     <p>The <code><![CDATA[<td-chips>]]></code> component has {{chipsAttrs.length}} properties:</p>
     <md-list>
--- a/src/app/components/components/chips/chips.component.html
+++ b/src/app/components/components/chips/chips.component.html
@@ -1,0 +1,8 @@
+<md-card>
+  <md-card-title>Chips</md-card-title>
+  <md-card-subtitle>Build a list of strings</md-card-subtitle>
+  <md-divider></md-divider>
+  <md-card-content>
+    <td-chips></td-chips>
+  </md-card-content>
+</md-card>
--- a/src/app/components/components/chips/chips.component.html
+++ b/src/app/components/components/chips/chips.component.html
@@ -4,21 +4,23 @@
   <md-divider></md-divider>
   <md-card-content>
     <p>Basic Demo</p>
-    <td-chips></td-chips>
+    <td-chips placeholder="Enter any string" [readOnly]="readOnly"></td-chips>
     <p>Autocomplete Demo</p>
-    <td-chips [items]="items"></td-chips>
+    <td-chips [items]="items" placeholder="Enter any string" [readOnly]="readOnly"></td-chips>
     <p>Autocomplete and requireMatch Demo</p>
-    <td-chips [items]="items" [(ngModel)]="itemsRequireMatch" requireMatch></td-chips>
-    <p>ReadOnly Demo</p>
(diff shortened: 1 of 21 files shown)


src/platform/chips/chip.component.scss
```diff
@@ -0,0 +1,18 @@
+:host {
+  display: block;
+}
+
+div {
```
kyleledbetter: why's this needed? targeting a div seems broad even for shadowDOM :)

kyleledbetter:
pushed up some SCSS updates to align closer to material2's way of styling children
also updated styles to use more variables and more closely match material design chip spec

scss-updates

src/platform/chips/chip.component.scss
emoralesb05: lol, well there was only one div haha but i guess i should've added a class :smile:

src/app/components/components/chips/chips.component.html
```diff
@@ -0,0 +1,78 @@
...
+  </md-card-actions>
+</md-card>
+<md-card>
+  <md-card-title>TdChipsComponent</md-card-title>
+  <md-card-subtitle>How to use this component</md-card-subtitle>
+  <md-divider></md-divider>
+  <md-card-content>
+    <h2><code><![CDATA[<td-chips>]]></code></h2>
+    <p>Use <code><![CDATA[<td-chips>]]></code> element to generate a list of strings as chips.</p>
+    <p>Add the [items] attribute to enable the autocomplete with a search list, and [requireMatch] to validated the input against the provided search list.</p>
```
emoralesb05: will do! thanks

src/platform/chips/chips.component.ts
```diff
@@ -0,0 +1,190 @@
...
+      this._length = this._value ? this._value.length : 0;
+    }
+  }
+  get value(): any { return this._value; };
+
+  ngDoCheck(): void {
+    // Throw onChange event only if array changes size.
+    if (this._value && this._value.length !== this._length) {
+      this._length = this._value.length;
+      this.onChange(this._value);
```
richavyas: Not sure why we have to use onChange here inside DoCheck?

src/app/components/components/chips/chips.component.html
```diff
@@ -0,0 +1,78 @@
...
+    <p>Basic Demo</p>
+    <td-chips placeholder="Enter any string" [readOnly]="readOnly"></td-chips>
+    <p>Autocomplete Demo</p>
+    <td-chips [items]="items" placeholder="Enter any string" [readOnly]="readOnly"></td-chips>
+    <p>Autocomplete and requireMatch Demo</p>
+    <td-chips [items]="items" [(ngModel)]="itemsRequireMatch" placeholder="Enter autocomplete strings" [readOnly]="readOnly" requireMatch></td-chips>
+    <p>Form Demo</p>
+    <form #form="ngForm">
+      <td-chips #spy [items]="items" [(ngModel)]="itemsForms" name="selections" placeholder="Enter autocomplete strings" [readOnly]="readOnly" requireMatch></td-chips>
+      <p>Dirty: {{form.form.dirty}}</p>
```
richavyas: Didn't understand the significance of this line.

src/platform/chips/chips.component.ts
emoralesb05: Its not the same as `ngOnChanges()`. Using onChange is something we need to do for the forms stuff and custom ControlValueAccesor for an array, so the form adds the `ng-dirty` class and removes the `ng-pristine` class when the value changes. And since its an array, its then the length changes at any point.

src/app/components/components/chips/chips.component.html
emoralesb05: It shows that we support forms to let the developer know when chips is dirty, pristine or touched.

richavyas: Cool. Either lets make it more explicit or remove it from the docs.

Issue #58, opened 2016-09-09
module(http): interceptors in ngModule


emoralesb05:
#### Feature Request

Change the way we provide interceptors in http module to be inline with how ngModule works. 

e.g.

`CovalentHttpModule.forRoot(interceptors);`

PR #62, opened 2016-09-11
Feature/new http

emoralesb05:
## Description

Update `http` module to provide interceptors to be inline on how ngModule works. 
https://github.com/Teradata/covalent/issues/58
https://github.com/Teradata/covalent/issues/65

### What's included?

- Removed depricated `provideInterceptors()` method from `http` module.
- Fixed bugs in interceptor pipeline where `Response`/`RequestOptions` objects passed through the pipeline would not be received at the end stage (success/error), even if the pipeline returned a different object.
- Polished code to be more maintainable in `http` module.
- Added `NgModule` notation and provided interceptors with `forRoot()`. e.g. `CovalentHttpModule.forRoot([InterceptorA, InterceptorB])`
- Updated `http` docs.
- Updated `README.md` with setup.

#### Test Steps

- [x] `ng serve`
- [x] Go to http://localhost:4200/#/components/http and see doc changes.
- [x] :boom:

Note: There is no easy way to test it, since you would have to create a service and interceptors just for the tests and point to a real API. (or mock API)

### DIFF
--- a/src/platform/http/http-interceptor.service.ts
+++ b/src/platform/http/http-interceptor.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, Type, ValueProvider, Injector } from '@angular/core';
+import { Injectable, Type, Injector } from '@angular/core';
 import { Http, RequestOptionsArgs, Response, Request } from '@angular/http';
 
 import { Observable } from 'rxjs/Observable';
@@ -121,19 +121,4 @@ export class HttpInterceptorService {
     });
   }
 
-}
-
-export function provideInterceptors(requestInterceptors: Type<any>[] = []): any[] {
-  let providers: any[] = [];
-  requestInterceptors.forEach((interceptor: Type<any>) => {
-    providers.push(interceptor);
-  });
-  providers.push({
-    provide: HttpInterceptorService,
-    useFactory: (http: Http, injector: Injector): HttpInterceptorService => {
-      return new HttpInterceptorService(http, injector, requestInterceptors);
-    },
-    deps: [Http, Injector],
-  });
-  return providers;
 }
--- a/src/platform/http/http-interceptor.service.ts
+++ b/src/platform/http/http-interceptor.service.ts
@@ -13,112 +13,79 @@ export interface IHttpInterceptor {
 @Injectable()
 export class HttpInterceptorService {
 
-  requestInterceptors: IHttpInterceptor[] = [];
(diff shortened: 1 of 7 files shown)


src/platform/http/index.ts
```diff
@@ -1,2 +1,35 @@
...
+
+@NgModule({
+  imports: [
+    HttpModule,
+  ],
+  providers: [
+    HttpInterceptorService,
+  ],
+})
+export class CovalentHttpModule {
```
ilsiepotamus: shouldn't this class definition be in it's own file?  typically i thought `index.ts` just described the things being exported from the different services.

src/platform/http/http-interceptor.service.ts
```diff
@@ -13,127 +13,79 @@ export interface IHttpInterceptor {
...
-    return this._http.request(url, options)
-      .do((response: Response) => {
-        return this._responseResolve(response);
-      }).catch((error: Response) => {
-        return this._errorResolve(error);
-      });
+    return this._setupRequest(this._http.request(url, this._requestResolve(options)));
   }
 
   delete(url: string, options: RequestOptionsArgs = {}): Observable<Response> {
```
ilsiepotamus: HTTP 1.1 allows for a body in a DELETE call- https://tools.ietf.org/html/draft-ietf-httpbis-p2-semantics-19

src/platform/http/http-interceptor.service.ts
```diff
@@ -13,127 +13,79 @@ export interface IHttpInterceptor {
...
-    this._requestConfig(options);
-    return this._http.delete(url, options)
-      .do((response: Response) => {
-        return this._responseResolve(response);
-      }).catch((error: Response) => {
-        return this._errorResolve(error);
-      });
+    return this._setupRequest(this._http.delete(url, this._requestResolve(options)));
```
ilsiepotamus: HTTP 1.1 allows for a body in a DELETE call- https://tools.ietf.org/html/draft-ietf-httpbis-p2-semantics-19

emoralesb05: Problem is that `Http` service in angular doesnt support a body parameter, we are just wrapping that service. Although for those cases, we have the generic `request` method, that allows to set a method DELETE with body.

ilsiepotamus: Maybe something we take note of and deal with later, esp. since angular doesn't support it.

PR #92, opened 2016-10-03
feature(data-table): new data-table component

fcoury:
## Description

This branch introduces a new Covalent component `td-data-table`. It was heavily inspired on Angular 1's *md-data-table*.

### What's included?

- Features:
   - [x] Basic table layout
   - [x] Automatic columns when none provided
   - [x] Row selection
   - [x] Multiple or single row selection
   - [x] Search
   - [x] Pagination
   - [x] Sorting
   - [ ] In line edit
- An incomplete example of its usage
- Updated documentation (WIP)

#### Test Steps

- Navigate to http://localhost:4200/#/components/data-table
- Check examples

datatable


kyleledbetter:
w00t thx @fcory! We'll jump in to test and collaborate

kyleledbetter:
nice job on the merge and the latest updates!

I really like the auto-focus on the input when you click the filter button

It would be nice if the x button to close the search would also cancel/empty the search input
image

fcoury:
Agreed @kyleledbetter and I'll implement that. I had a blocker but I'll work on cleaning it today.

Meanwhile, I have another blocker for inline editing, since the dialog module for Angular 2 Material is not finished.

Do you have any ideas on how we could overcome this? 

/cc @emoralesb05

fcoury:
Done @kyleledbetter. I could still use some guidance on how to proceed with the inline edit.

fcoury:
A few things I wanted to discuss/implement next:

## Sorting

This is the low hanging fruit, as I don't have any blockers and it seems I don't need any input here.

## Inline editing

The problem here is that we need a working dialog module to make Inline Editing work the way it's spec'ed out [here](https://material.google.com/components/data-tables.html):

Material Data Table Inline Editing

## Formatting / Alignment

Should we include formatting and alignment as part of the column metadata? What would you suggest we use to format numbers/date/etc.?

A few libs we used in the past:

- [Moment.js](http://momentjs.com/) for date formatting
- [Numeral.js](http://numeraljs.com/) for numeric formatting

## Events

I think we should trigger a few events:

- Row(s) selected (passing an array of selected rows for the case where all rows are selected)
- Row(s) unselected (same as above)
- Column changed (when Inlined Editing is working) -- this event can return `true`/`false` to indicate whether the change is accepted or not

What else am I missing?

fcoury:
OK, sorting done:

sorting demo

kyleledbetter:
tiny comment on the sorting icon: in the spec it's on the left side of the column heading and it's a slightly different icon `arrow_downward` and `arrow_upward`:

image

src/platform/data-table/data-table.component.html
```diff
@@ -0,0 +1,78 @@
...
+          </md-checkbox>
+        </th>
+        <th td-column *ngFor="let column of columns" [ngClass]="{'md-clickable': _sorting}" (click)="setSorting(column)">
+          <span class="md-caption">
+            {{column.label}}
+            <md-icon *ngIf="column === _sortBy && _sortOrder === _ascending">
+              keyboard_arrow_down
+            </md-icon>
+            <md-icon *ngIf="column === _sortBy && _sortOrder === _descending">
+              keyboard_arrow_up
```
kyleledbetter: `arrow_downward` here

aciccarello:
@kyleledbetter The spec also says the text alignment depends on the data type:

>**Text alignment**
>Right-aligned numeric columns
>Left-aligned text columns

Could that change what side the ascending/descending icon appears on?

kyleledbetter:
@fcoury for dates and numeric formatting do you think you can use the Covalent Pipes we have available? https://teradata.github.io/covalent/#/components/pipes

we have
- digits (most of your numeric conversion like: Digits: 4521 Converted: 4.5 K)
- byte conversion (filesize like Bytes: 72452903343 Converted: 72.45 GB)
- time ago (like 3 months ago, 2 hours ago, etc)
- time difference (hours/days/etc between 2 dates)
- truncate (truncate to whatever length and add ... )

kyleledbetter:
@aciccarello great question, i'd think the sort arrow icon location is more a question of LTR vs RTL than the numeric/text alignment, like it would always be on the left for english

kyleledbetter:
@fcoury for the Inline Editing I think you can tap into the overlay services available now (which are used by dialogs which are also working in `develop` branch you're working off https://github.com/Teradata/covalent/blob/develop/src/app/components/style-guide/material-components/material-components.component.ts#L124-L183 )

I'm guessing you'll have to tap into overlay https://github.com/angular/material2/tree/master/src/lib/core/overlay & portal https://github.com/angular/material2/tree/master/src/lib/core/portal

kyleledbetter:
@fcoury another key feature you can add is the Alternative Header when you select items

image

fcoury:
@kyleledbetter @aciccarello thanks for all the comments, I'll be going through them tomorrow. More to come... Thanks! 😀

src/platform/data-table/data-table.component.html
fcoury: Thanks, adding both!

fcoury:
@kyleledbetter @aciccarello @emoralesb05 just added formatting support and made the sorting look and feel match md-data-table:

data-table sorting

kyleledbetter:
Latest looks awesome!

Couple of requests:

- [ ] option make the whole row clickable when the select box is enabled
- [ ] when the whole row is clickable for select add the md-ripple attr (something like `[attr.md-ripple]="_rowSelection === true"`)

fcoury:
@kyleledbetter where do we add the `md-ripple` effect? I tried adding it to both the `table` and `tr` elements, made sure it worked, but but nothing changed.

kyleledbetter:
@fcoury yeah i just tested and had a hard time getting it to work properly. I can get it working on certain elements if i just put `md-ripple` on them... need to do more testing on this one

emoralesb05:
@kyleledbetter @fcoury The reason `md-ripple` isnt working its because you need to import `MdRippleModule` where you want to use it. So to avoid that, you can just import `MaterialModule` and let `webpack`/`rollup` do the necessary tree shaking.

emoralesb05:
Going over some of the comments to see what has been discussed. 

## Events

So for events i think a simple `(change)` output would work fine with a `IDataTableEvent` that has the `source` element (this) + an array of the selected indexes.
```
interface IDataTableEvent {
  source: TdDataTableComponent;
  values: any[]; // this is a list of rows or something that specifies which ones are selected.
}
```

## Inline editing

We can try and use a `prompt` dialog that we have in the `simple dialogs` feature and leverage it if needed. Simple as using `dialogsService.openPrompt({value: 'value'});` and it uses the second dialog in the `data-table` spec.

image

Although we might need to add a few things like `character count` and what not.

## Tooltip description

Would be cool to be able to add an optional `tooltip` description as part of the column metadata to follow the `data-table` specs.

image

```
export interface TdDataTableColumn { 
  name: string, 
  label: string,
  tooltip?: string,
  numeric?: boolean,
  format?: { (value: any): any };
};
```

## Paging

Later on it would be nice to have a `td-paging` component that gets hooked into `td-data-table` like this: 

```
<td-data-table>
  <td-paging (next)="next()" (prev)="prev()" (change)="onPagingChange($event)"><td-paging>
</td-data-table>
```

That way we can hook it into `td-data-table` or us it stand alone with `md-list` or w/e we want. I already have the idea in my mind, just setting up a brainstorm flow for a second iteration of `data-table` haha


We need to specify a goal for the first iteration and start polishing the base rather than adding more and more things for now. This is awesome work though! :boom:

emoralesb05:
Btw, do we really need ripple? the spec doesnt say anything about it.. and `<table>` layout just screws ups the ripple since ripple uses absolute positioning.

kyleledbetter:
@emoralesb05 it's a great question, we don't have to have ripple as it's not in the spec, it just felt like a natural extension of the spec :)

As long as we have the proper row hover and selected bg colors I'm happy

src/platform/data-table/_data-table-theme.scss
```diff
@@ -0,0 +1,62 @@
+@import '~@angular/material/core/theming/palette';
+@import '~@angular/material/core/theming/theming';
+
+// TODO properly wrap in mixin to apply theme like material2 components
+$foreground: $md-light-theme-foreground;
+$background: $md-light-theme-background;
+$accent: md-palette($md-light-blue, 50);;
```
kyleledbetter: just noticed this line has an extra `;`

kyleledbetter:
Just tested the latest and this UX looks great for the first release @fcoury ! Pls get everything passing tests and ready for release!

src/platform/data-table/README.md
```diff
@@ -0,0 +1,43 @@
+# td-chips
```
emoralesb05: We need to update the read me for this since its using the `td-chips` base.

src/platform/data-table/data-table.component.html
```diff
@@ -0,0 +1,101 @@
...
+  </form>
+  <button md-icon-button (click)="toggleSearch()">
+    <md-icon>close</md-icon>
+  </button>
+</div>
+
+<div class="md-table-container" *ngIf="_visibleData.length > 0" title>
+  <table td-table
+         [class.md-row-select]="_rowSelection"
+         [attr.md-ripple]="_rowSelection === true">
```
emoralesb05: This isnt really doing anything at all, and since we arent gonna do anything with ripple, we can remove it. 
Its just adding a random attribute named `md-ripple` without actually compiling it into a directive (since its done after the ng2 compiler renders everything).

src/platform/data-table/data-table.component.ts
```diff
@@ -0,0 +1,330 @@
...
+
+export interface ITdDataTableColumn {
+  name: string;
+  label: string;
+  tooltip?: string;
+  numeric?: boolean;
+  format?: { (value: any): any };
+};
+
+export interface ITdDataTableSortChanged {
```
emoralesb05: Can we maybe change the name of this to `ITdDataTableSortEvent`?

fcoury:
@emoralesb05 I think I addressed all your feedback. We are only missing the lodash removal. I was basically waiting instruction on how you guys approach generic/utility methods that could be reused throughout the framework.

kyleledbetter:
The is a wonderful first pass at datatables, fantastic work all!
I'm good to merge, would like one more sign off on the code real quick

emoralesb05:
Everything looks great!

totalard:
Any Future Update on Column Filter. It is a Much Needed Feature. Like this http://codepen.io/iamisti/pen/kkvBxd

webplease:
Hi to all,

when set [clickable]="true" all the row is obviously clickable ... but if I add a button in a row-cell the click event of the button is ignored ... is there a way to solve?

PR #97, opened 2016-10-11
Added auto trim directive for string values

emoralesb05:
## Description

`TdAutoTrim` to trim string values on `blur` event. `ngModel` is required since it hooks into it.

#### Test Steps

- [x] use anywhere in code `<md-input tdAutoTrim [(ngModel)]="val"></md-input>`

### DIFF
--- a/src/app/components/components/components.component.ts
+++ b/src/app/components/components/components.component.ts
@@ -68,6 +68,11 @@ export class ComponentsComponent {
     route: 'http',
     title: 'Http',
   }, {
+    description: 'Core directives &amp; utilities',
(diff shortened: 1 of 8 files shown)


kyleledbetter:
should we document this anywhere or maybe create a Core/Utilities page in the /components area of docs?

emoralesb05:
Yes, it might be a good idea to do so. We can merge this and validations and then create a PR to add docs for every directive/validator.

kyleledbetter:
i'm about to push to this branch so we can get it started

kyleledbetter:
image

PR #116, opened 2016-10-25
Changes to support multi-column sort.

thomaslennan:
## Description

Promoted sortable property from the table and only one column to column metadata to allow sorting on multiple columns. Enhanced docs to describe table and column properties.

### What's included?

- Modified core data table component.
- Updated sample data table component to utilizing new functionality.
- Updated README.md to document new functionality.

#### Test Steps

- Launch app, navigate to the data table sample page.
- Click on column headers and notice that only the first three columns are sortable.
- Click on column headers and confirm entire table is sorted by the current column values.

##### Screenshot

datatablemulticolumnsort

### DIFF
--- a/src/app/components/components/data-table/data-table.component.html
+++ b/src/app/components/components/data-table/data-table.component.html
@@ -6,53 +6,22 @@
     <p>Basic Demo</p>
 
     <td-data-table
+      title="Search and pagination enabled"
       [data]="data"
       [columns]="columns"
       [pagination]="pagination"
       [pageSize]="pageSize"
       [rowSelection]="rowSelection"
-      [multiple]="multiple"
-      [sorting]="sorting"
-      [sortBy]="sortBy"
+      [multiSelect]="multiSelect"
+      [search]="search"
+      [initialSortColumn]="initialSortColumn"
       [sortOrder]="sortOrder"
-      search="true"
-      title="Search and pagination enabled"
       (sortChanged)="sortChanged($event)">
     </td-data-table>
 
   </md-card-content>
   <md-divider></md-divider>
   <md-card-actions>
-    <button md-button color="primary" (click)="toggleSorting()">
-      Toggle Sorting
-      ({{sorting ? 'ON' : 'OFF'}})
-    </button>
-
-    <button md-button color="primary" (click)="toggleSortBy()" [disabled]="!sorting">
-      Toggle SortBy
-      ({{sortBy}})
-    </button>
-
-    <button md-button color="primary" (click)="toggleSortOrder()" [disabled]="!sorting">
-      Toggle SortOrder
-      ({{sortOrder}})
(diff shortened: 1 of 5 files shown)


src/app/components/components/data-table/data-table.component.html
```diff
@@ -6,62 +6,51 @@
     <p>Basic Demo</p>
 
     <td-data-table
+      title="Search and pagination enabled"
```
kyleledbetter: just curious, why remove the title? i thought it was broken :)

src/app/components/components/data-table/data-table.component.html
```diff
@@ -6,62 +6,51 @@
...
       [data]="data"
       [columns]="columns"
       [pagination]="pagination"
       [pageSize]="pageSize"
       [rowSelection]="rowSelection"
-      [multiple]="multiple"
-      [sorting]="sorting"
-      [sortBy]="sortBy"
+      [multiSelect]="multiSelect"
```
kyleledbetter: all these changes to attribute names result in breaking changes, was there a reason beyond preference?

kyleledbetter:
multi
Hi @thomaslennan I'm a little confused by the UX of what's happening. When i click multiple column headers, i can't keep the asc/desc orders from one to the other. I tried holding Shift or Cmd.

I tested the multi-sort in PrimeNG which uses Cmd to multi select and it seemed to work most like what i expected, but i honestly have a hard time wrapping my head around the value, it's confusing to sort by more than one column for me...
multi-prime

src/app/components/components/data-table/data-table.component.html
thomaslennan: Not sure I understand. I added the title property to try and be consistent.

src/app/components/components/data-table/data-table.component.html
emoralesb05: The reason `multiple` was selected, its because it goes inline with how the native ` multiple` attr works in html elements. If you are used to `HTML5`, `multiple` is intuitive.

e.g.
`<input type="file" multiple>`

aciccarello:
I feel that only marking the primary sorted column would be acceptable so long as a stable sorting algorithm is used.

src/app/components/components/data-table/data-table.component.html
thomaslennan: Ah, that makes good sense.

thomaslennan:
Hum, I think I misconstrued how sorting was working on the original code base. I missed the code in setSorting that altered the sort column on click of the column header. Seems like the original sort functionality is actually what I would want (all columns are sortable up and down). Seems like the naming of the sortBy property is a bit misleading.

And, Kyle I wasn't intending for column sorts to be additive. Just that more than one column could be the subject of sorting.

Issue #114, opened 2016-10-22
Create paging component


emoralesb05:
#### Feature Request

Create a paging component so it can be used in lists or data-table (abstract it from data-table).

#### What is the expected behavior?

Use `<td-paging-bar>` to render a paging component.

#### What is the motivation / use case for changing the behavior?

Being able to add it anywhere where paging is needed.

emoralesb05:
Closed with https://github.com/Teradata/covalent/pull/117

PR #117, opened 2016-10-25
Feature/paging-bar

emoralesb05:
## Description

Created standalone paging-bar component to be flexible for any use case following the specs in https://material.google.com/components/data-tables.html#data-tables-specs
https://github.com/Teradata/covalent/issues/114

#### `<td-paging-bar>`

Generate a paging bar.
The (change) event throws an event with the following interface:
```typescript
export interface IPageChangeEvent {
  page: number;
  maxPage: number;
  pageSize: number;
  total: number;
  fromRow: number;
  toRow: number;
}
```
The `<td-paging-bar>` component has 5 properties:
image

Usage in HTML:
```html
<td-paging-bar [pageSizes]="[100,200,500,1000,2000]" [pageSize]="100" [total]="1345" (change)="change($event)">
</td-paging-bar>
```

#### Test Steps

- [x] `ng serve`
- [x] Go to http://localhost:4200/#/components/paging-bar
- [x] Play with demo.

image
td-paging-bar-demo

### DIFF
--- a/src/platform/paging/paging-bar.component.ts
+++ b/src/platform/paging/paging-bar.component.ts
@@ -63,6 +63,10 @@ export class TdPagingBarComponent implements OnInit {
     return this._page;
   }
 
+  get maxPage(): number {
+    return Math.ceil(this._total / this._pageSize);
+  }
+
   @Output('change') change: EventEmitter<IPageChangeEvent> = new EventEmitter<IPageChangeEvent>();
 
   ngOnInit(): void {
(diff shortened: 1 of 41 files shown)


kyleledbetter:
nice! should we allow a selection of "All" pages in the menu?
image

kyleledbetter:
Optimized SCSS and split out theme:

image

kyleledbetter:
## Updates 

- [ ] make first & last page buttons optional
- [ ] add "All" in pageSizeMenu if possible

src/platform/paging/paging-bar.component.ts
```diff
@@ -0,0 +1,180 @@
...
+  private _toRow: number = 1;
+  private _initialized: boolean = false;
+
+  /**
+   * pageSizes?: number[]
+   * Array that populates page size menu. Defaults to [50, 100, 200, 500, 1000]
+   */
+  @Input('pageSizes')
+  set pageSizes(pageSizes: number[]) {
+    if (pageSizes instanceof Array) {
```
emoralesb05: Yeah, will add an error message if its not a valid input :+1:

emoralesb05:
I think `all` rows defeats the purpose of a paging component. 
The consumers can just set a really big number if they realllyyy wanna do that.

kyleledbetter:
The code would have to be changed to do that, while the end user would be the one to select All. It's a ux thing if you're not wanting to go through all pages to view everything and very nice to have

emoralesb05:
I guess what im trying to say is, if they wanna see all rows they could either turn off paging or add the total as pageSize option.
e.g.
`[pageSizes]="[100,200,500,1000, data.length]"`

kyleledbetter:
`[pageSizes]="[100,200,500,1000, data.length]"` still puts it in the developers hands, not the end user. The end user would often want to just display All

emoralesb05:
Right, but its up to the developer to do something with that `all` either way, so the developer still need to do it.

emoralesb05:
Alright, you convinced me. Optional, default to false. lol

emoralesb05:
- [x] added `pageSizeAll` input
```
<td-paging-bar [firstLast]="false" [pageSizeAll]="true" [pageSizes]="[100,200,500,1000,2000]" [pageSize]="100" [total]="1345" (change)="change($event)">
</td-paging-bar>
```
Output:
image

PR #127, opened 2016-10-28
Rename title attribute on layouts

mrjayhawk:
## Description

Renamed the title attribute on layouts to instead be sidenavTitle and toolbarTitle instead of title and updated title and subTitle on td-layout-card-over to be cardTitle and cardSubTitle.  Updated all the documentation and ReadMe to reflect these new changes.  The attribute title was causing a tooltip on hover anywhere on the page to show the value passed in.  Renaming this removes the clashing with the native title attribute.
### What's included?

Changes to html and typescript files as well as one test spec file and documentation and readme files
#### Test Steps
- [x] make sure pages that use layouts still show the title but no hover tooltip

### DIFF
--- a/src/platform/core/layout/layout-nav-list/layout-nav-list.component.ts
+++ b/src/platform/core/layout/layout-nav-list/layout-nav-list.component.ts
@@ -18,6 +18,7 @@ export class TdLayoutNavListComponent {
   /**
    * title in toolbar
    */
+  @Input('title') title: string; // deprecated
   @Input('toolbarTitle') toolbarTitle: string;
 
   /**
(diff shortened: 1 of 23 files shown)


src/platform/core/layout/layout-nav-list/layout-nav-list.component.ts
```diff
@@ -18,7 +18,7 @@ export class TdLayoutNavListComponent {
   /**
    * title in toolbar
    */
-  @Input('title') title: string;
+  @Input('toolbarTitle') toolbarTitle: string;
```
kyleledbetter: just thinking out load for backwards compatibility:
what if we did:
```
@Input('title') toolbarTitle: string;
@Input('toolbarTitle') toolbarTitle: string;
```

or whatever the proper approach would be here so we could keep the backwards compat w/ `title` and set it to `toolbarTitle`

PR #146, opened 2016-11-10
Typofix

BernhardRode:
To simple =)


kyleledbetter:
ha, good catch @BernhardRode thx :)

PR #162, opened 2016-11-20
MINOR: Missing debounceTime operator

### LINKED ISSUE #161: [Comp: Search] debounceTime is missing as a observable method
There is a possible bug in the to be released version. Since Rollup is used for the bundling of the build, the debounceTime operator which is used in the search component is not included in the built 

I will submit a pull request to fix it.

asaph26:
Added debounceTime operator to the rollup file. Fixes #161


emoralesb05:
Awesome! thanks!

PR #186, opened 2016-12-12
feat(animations): added collapse animation and used in steps and expansion

emoralesb05:
## Description

Added animation `TdCollapseAnimation` and able to reuse it.

#### Usage

typescript
```
import { TdCollapseAnimation } from '@covalent/core';
...
...
animations: [
  TdCollapseAnimation(200),
],
```

html
```
<div [@tdCollapse]="true|false">
  ...
</div>
```

The reason is because `TdToggleDirective` had many issues executing the animations with the `AnimationPlayer` class and its better to do animations the [ng2] way either way. 

### What's included?

- Fixed bug with overlapping content when animating stepper and expansion in FF50.
- [TdCollapseAnimation] animation.
- Replaced `TdToggleDirective` usage in steps and expansion-panel with `TdCollapseAnimation`

#### Test Steps

- [ ] `ng serve`
- [ ] Go to demo in stepper or expansion-panel.
- [ ] Check it works fine in FF and Chrome.

### DIFF
--- a/src/platform/core/animations/collapse.animation.ts
+++ b/src/platform/core/animations/collapse.animation.ts
@@ -15,10 +15,12 @@ export function TdCollapseAnimation(duration: number = 150): AnimationEntryMetad
     state('true', style({
       height: '0',
       overflow: 'hidden',
+      display: 'none',
     })),
     state('false',  style({
       height: '*',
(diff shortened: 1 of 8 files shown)


richavyas:
LGTM

Issue #187, opened 2016-12-12
Filtering does not work in data-table


emoralesb05:
#### Bug Report

Filtering does not work in data-table when providing a format function in the column object.

#### What is the expected behavior?

Should filter accordingly to the format function.

emoralesb05:
Fixed and will be available in 0.10

PR #188, opened 2016-12-13
fix(data-table): format columns correctly when displaying them. closes(#187)

emoralesb05:
## Description

Switched the way we format the column values in the `data-table` component to be on display, so the underlying data is not modified nor copied.
https://github.com/Teradata/covalent/issues/187

#### Test Steps

- [x] `ng serve`
- [x] Go to http://localhost:4200/#/components/data-table
- [x] Play around with column formats in the demos.

image


emoralesb05:
- [x] Added validation to properly handle empty (undefined) data

kyleledbetter:
Does the example HTML & TS need updating also?

image

emoralesb05:
Well, we didnt need to, but i guess it doesnt hurt to show how they can do it with the atomic components :stuck_out_tongue: pushed~

PR #217, opened 2016-12-28
feature(paging): usage of md-select as menu

emoralesb05:
## Description

Replaced `md-icon-buton` and `md-menu` with `md-select`


kyleledbetter:
updated styles and made width for select dynamic
image

PR #243, opened 2017-01-09
bugfix(docs): theme styles duplication

emoralesb05:
## Description

Since `theme.scss` was included in the `angular-cli.json` and as part of an `import` in the `styles.scss`, the file was actually getting included twice. This meant duplicated `css` rules and a bigger file.

#### Test Steps

- [x] `ng serve` with out the fix and check `styles` kb size. (`559kb`)
image
- [x] `ng serve` with the fix and check `styles` kb size. (`473kb`)
image

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -1,18 +1,4 @@
-@import "theme";
 // You can add global styles to this file that will cascade like normal CSS
-
-// Active icon color in list nav
-md-nav-list {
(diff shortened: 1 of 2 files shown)


indyplanets:
@emoralesb05 LGTM!

screen shot 2017-01-09 at 12 49 48 pm

richavyas:
Checked pages, LGTM.

PR #245, opened 2017-01-11
bugfix(md-icon): fix alignment issues

HitomiTenshi:
## Description

This PR fixes alignment issues when using smaller icon size classes (e.g. `.material-icons.md-18`) than the default one (`.material-icons.md-24`).

### What's included?

- Changes to `width` and `height` in `src/platform/core/common/styles/font/_font.scss`
- Added icon size section to `src/app/components/components/material-components/material-components.component.html`

#### Test Steps

Using `<md-icon class="md-18">filter_none</md-icon>` in an icon-button without this fix:

electron_2017-01-11_12-14-34

Using `<md-icon class="md-18">filter_none</md-icon>` in an icon-button **with** this fix:

electron_2017-01-11_12-16-50


emoralesb05:
Awesome! We will test this and merge afterwards. Thanks :tada:

HitomiTenshi:
Oh sorry, I didn't compile and check after doing the changes. My bad.

You guys are already using the `.md-icon-font-xl` classes on those icons, but you do not have any CSS defined for them.

I'll fix that.

HitomiTenshi:
I put the width and height changes into ´src/platform/core/common/styles/_icons.scss´, since `.md-icon-font-xl` is a class specific to covalent and not the official Material Icon CSS.

@indyplanets I went through the site this time and everything seems to be in order now. 👍

indyplanets:
@HitomiTenshi Thanks for those changes! 👍 

I've been testing this out and making sure we cover everything here. Looks like this change has some unwanted consequences. See screenshot:
md-icon

Might be a better option to add specific `width` and `height` styles for each `md-18`, `md-24`, etc... classes.

HitomiTenshi:
Alright, I have been going at this all wrong. While I was searching for the cause of this dilemma, I noticed that the icons still received `width` and `height` from somewhere. That's when I found [material 2's icon definitions](https://github.com/angular/material2/blob/master/src/lib/icon/icon.scss).

The material 2 devs add width and height definitions to `md-icon`, which seems to cause the misalignment issues on icons that have a font-size smaller than 24px.

The solution to that was simple: Replace the current definition of `1em` with `auto` on `width` and `height`, which will override material 2's `md-icon` definitions and automatically apply correct sizes when using smaller icons.

With `auto` your `md-card-colored` icons resize and align appropriately as well without the need of extra css.

I went ahead and added an Icon Size section to the docs, so you can see the effect. I also fixed a few alignment issues there while I was at it. 😎

src/app/components/components/material-components/material-components.component.html
```diff
@@ -205,85 +205,121 @@
...
+      <div flex="50" layout="column">
         <h3>Icons with Buttons</h3>
-        <button md-icon-button>
+        <div layout="row" layout-align="start center">
+          <button md-icon-button>
             <md-icon>favorite</md-icon>
-        </button>
-        <button md-fab color="accent">
+          </button>
+          <button md-fab color="accent" style="margin-left: 10px;">
```
kyleledbetter: these inline styles are just for demo spacing right? we strive (but don't always achieve for the demos to match the code example, and we obviously wouldn't want to recommend inline styles. maybe put these buttons in a md-toolbar, or a `<div layout="row" layout-align="space-between center">` for the demo so they have proper spacing?

HitomiTenshi: Well the dividers had inline styles and the radio button demo also had `style="padding-right:10px;"`, so I thought that you guys were okay with it, since it's a demo and all.

I could change the `layout-align` to `"space-between center"` if you want, but since pretty much everything else on that particular page is kinda inline styled, the page would need a re-edit someday anyway.

emoralesb05: For the most part the `material components` demo docs need to be divided so its not `1` big component with proper layouts. 

For this PR in particular, i would not make too much emphasis on inline styling since we do it everywhere else in that view, and just focus on the fix and display of the different fonts (which its there now)

indyplanets:
@HitomiTenshi 
screen shot 2017-01-13 at 3 43 12 pm
Thanks for adding that section! 👍 🦄 
PR approved for merge.

kyleledbetter:
thx for the fix @HitomiTenshi! 🍺

PR #259, opened 2017-01-17
feature(loading): loading enhancements + re-haul to use material modules. (closes #255) (closes #84)

### LINKED ISSUE #84: [Feature Request] Loading Overlay Component Target
#### Feature Request

This is a use enhancement for the current Loading Indicator Overlay to extend the API.

Currently, there are two methods to invoke the creation of a Loading Indicator Component, each which have different outcomes.

The Template Syntax method replaces everything within its scope with a Loading Component where the Loading Service call appends an Overlay the current view obscuring the content though not completely.

I would like to see a way in which the Overlay can be created with the same constraints as the Template Syntax method so that you can target which component the Overlay is obscuring.

The Loading Service requires a ViewContainerRef be passed into it, however it only accepts Root references, meaning that the following cannot be utilized.

`@ViewChild(TargetComponent, {read: ViewContainerRef}) TargetComponent: TargetComponentType;`

The latter will give you an instance of ViewContainerRef, but it is not a complete instance, and therefore cannot be used with the service.

The only way to target a component for the overlay is to have the service as part of the targeted component. I want to achieve the creation of two Overlay loads, a PageLoader and a ScreenLoader, essentially. The PageLoader will Overlay only the Main Body Content (Not the Sidebar or the Header) which signifies that the user can navigate away from the page, and the action will not be canceled. The ScreenLoader will Overlay the Main Body Content and the Sidebar, to prompt the user that the action cannot be navigated from.

In my situation, I am utilizing a combination of layouts eventually drilling down to a single router-outlet  (similar as to how it is presented in the repo) for all page routes, which allows for a simple folder structure. As it stands, the only way for me to achieve my desired outcome would be to break up the main template into a few smaller templates and create the overlays within the service at the level of the individual templates which would get pretty messy.

I would like to suggest either that the Template Syntax is changed so that it does not replace the inner content, but provides the reference for creating an overlay for the inner content, or the Loading Service Construct is modified so that it utilizes a ShadowDOM reference rather than a root ViewContainerRef.

### LINKED ISSUE #255: color="primary" is hard coded in td-loading component
# Feature Request
Currently `color="primary"` is hard coded in `td-loading` component. We should make it dynamic so that developers can easily change progress-bar color to `accent` or `warn`.  By default, we can make it primary.

`color` property should be configurable

I want to show the progress-bar color as `accent` instead of `primary`

emoralesb05:
## Description

Code re-haul of loading.module for enhancements and better usage + unit tests + better demos.
https://github.com/Teradata/covalent/issues/255
https://github.com/Teradata/covalent/issues/84

### What's included?

- feature(animations): Added `tdFadeInOut` animation and used it instead of `tdFade` directive.
- feature(loading): Easier fullscreen loading creation + default fullscreen mask.
- feature(loading): Better efficiency with `OnPush` strategy.
- `Deprecate` `ILoadingOptions` in favor of `ITdLoadingConfig`
- `Deprecate` `createOverlayComponent()` method from `TdDialogService` in favor of `create()`
- `Deprecate` `loadingType` and `loadingMode` in favor of `tdLoadingType` and `tdLoadingMode` so the inputs can be used correctly in (*) syntax.    e.g. `*tdLoading="'name'; type:'linear'"`
- refactor(loading): Leverage `TemplatePortal` and `Overlay` from `material`.
- refactor(loading): separated code into creation (factory) and execution (service), and made `TdLoadingFactory` for internal usage only.
- feature(loading): Added `tdLoadingStrategy` to change between `overlay` and `replace` strat. https://github.com/Teradata/covalent/issues/84
- feature(loading): Added `tdLoadingColor` input for `primary` , `accent` or `warn` https://github.com/Teradata/covalent/issues/255
- internal(loading): Better error messages.
- internal(loading): More code blocks in methods.
- docs(loading): Better demo and examples.
- code-health(loading): unit tests for `TdLoadingDirective` and `TdLoadingService`

#### Test Steps

- [x] `ng serve`
- [x] go to http://localhost:4200/#/components/loading and see new demo and examples.
- [x] `npm run tslint`
- [x] `npm test` and see code coverage now.
- [x] `npm run build`

Before unit tests:

image

image

After unit tests:

image

image

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ install:
 - npm i process-nextick-args util-deprecate buffer-shims
 - npm install
 script:
-- npm run tslint
+- npm run lint
 - npm run test
 after_script:
 - npm run coveralls
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -11,7 +11,7 @@ Execute `npm run start-release -- [version]` to start the automatic release proc
   1. Creating a `release/v[version]` branch using `git flow release`.
   2. Bumping its version to [version] release and commiting bumped version files.
   3. Publishing `release/v[version]` branch into repository. 
-  4. Executes `npm run tslint` and `npm run test`.
+  4. Executes `npm run lint` and `npm run test`.
 
 The release is published in case there is a need for any additional tests, version fixes or bug fixes. This can be added before it is actually released.
 
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "scripts": {
     "e2e": "protractor",
     "e2e-test": "protractor ./protractor.conf.js",
-    "tslint": "tslint -c ./tslint.json \"./src/**/*.ts\" -e \"./src/**/typings.d.ts\" -e \"./src/environments/**\"",
+    "lint": "tslint -c ./tslint.json \"./src/**/*.ts\" -e \"./src/**/typings.d.ts\" -e \"./src/environments/**\"",
     "postinstall": "webdriver-manager update",
     "webdriver-update": "bash ./node_modules/.bin/webdriver-manager update",
     "test": "ng test --code-coverage --single-run",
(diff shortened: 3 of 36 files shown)


kyleledbetter:
It would be nice if:

- [ ] the top demo had some loader running by default
- [ ] the `Register` buttons read `Toggle Loader`

is the breaking change: Renamed ILoadingOptions into ITdLoadingConfig absolutely necessary? :)

Issue #260, opened 2017-01-17
[Component] Data-List (responsive list version of data-table)


richavyas:
### Description
Data List is a feature for Managing a data list with various sub-features like: 
  - list
  - filter
  - order by
  - search
  - pagination
  - responsiveness

#### Filter
User optionally might want to see values for some columns and this filter gives a great way of enabling that e.g. various date options for a list `created/updated/last opened` can be clubbed into one column using this filter and the column header can be rendered as a select box for those date columns.

#### Order By
On the click of `asc/desc` icon on a column, the list can be ordered by.

`<td-data-list>` is a tag for creating List. It has the following sub-tags for columns and various features:
  - `<td-data-list-column>`
  - `<td-data-list-search>`
  - `<td-data-list-pagination>`

##### `<td-data-list>`
List tag consists of the following parameters:
  - list title
  - optional icon for the list
  - dynamic data to be rendered by the list

##### `<td-data-list-column>`
Column tag consists of the following parameters:
  - name
  - icon : optional
  - flex : value for the column layout
  - columnOptions : comma separated list of options for filter-select-box
  - sort : if sorting is required on this column

##### `<td-data-list-search>`
List-search tag consists of the following parameters:
  -  columns : columns to be searched

##### `<td-data-list-pagination>`
List-pagination tag consists of the following parameters:
  - pageSizes : record counts per page
  - totalRecords : total number of records

Usage:

```
<td-data-list
        [listTitle]="[string]"
        [listIcon]="[icon from Covalent icons]"
        [data]="[data to be rendered]">

        <td-data-list-column [name]="[string]" [icon]="[icon from Covalent icons]" [flex]="[number]"
                [columnOptions]="[comma separated list of options for filter-select-box]" [sort]="[true]">
        </td-data-list-column>

        <td-data-list-search [columns]="[columns to be searched]"></td-data-list-search>

        <td-data-list-pagination 
                [pageSizes]="[number of records to be shown during pagination]" 
                [total]="[maximum number of records to be rendered on a page]">
        </td-data-list-pagination>

</td-data-list>
```

![data-list](https://cloud.githubusercontent.com/assets/4469548/22125319/0e6aa7bc-de48-11e6-8d4d-9c036b26b552.gif)

emoralesb05:
So, just to keep start the flow here:

### Component reuse and flexibility

[...]

PR #262, opened 2017-01-17
feature(Management List): A new design pattern for sorting/filtering items #260

richavyas:
## Description

Management List, a new design pattern for sorting/filtering items. Takes care of #260 

### What's included?
- Management List with sorting & filtering design pattern. 
- Filter options to be added as one of the column headers and the list can be filtered based on what option has been selected. 
- Based on the selected filter option, list can also be sorted say `asc/desc`.

#### Test Steps

- [x] ng serve
- [x] then navigate to http://localhost:4200/#/style-guide/management-list
- [x] 🎉 

filter-sort-menu


kyleledbetter:
I'm not sure if design patterns and non-platform code needs unit tests

src/platform/core/management-list/management-list.module.ts
```diff
@@ -0,0 +1,28 @@
...
+    CovalentCommonModule.forRoot(),
+  ],
+  declarations: [
+    TdManagementListComponent,
+  ],
+  exports: [
+    TdManagementListComponent,
+  ],
+})
+export class CovalentManagementListModule {
```
emoralesb05: Im not sure we want to add a module just for a design pattern. The whole code should be in the design pattern page.

src/platform/core/management-list/management-list.ts
```diff
@@ -0,0 +1,41 @@
...
+  [key: string]: OrderBy;
+}
+
+@Component({
+  selector: 'td-management-list',
+  templateUrl: './management-list.html',
+  styleUrls: ['./management-list.scss'],
+})
+
+export class TdManagementListComponent {
```
emoralesb05: Any reason this is a component and not a service?

src/app/components/style-guide/management-list/management-list.component.html
```diff
@@ -0,0 +1,94 @@
...
+  <md-card-content>
+    <div class="md-padding bgc-grey-200">
+      <md-card>
+        <md-list class="pull-top">
+          <md-list-item>
+            <md-icon md-list-avatar>apps</md-icon>
+            <h3 md-line>App Name</h3>
+            <span flex></span>
+            <span flex="35" class="pad-left md-body-1 tc-grey-500">Owner</span>
+            <span class="md-sort-header" flex="40" layout="row" layout-align="center center" class="pad-left md-body-1 tc-grey-500">
```
emoralesb05: We can probably leverage the `md-select` now to do this.

src/app/components/style-guide/management-list/management-list.component.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  dateOptions: Object[] = ['Updated', 'Created'];
+  dateFilter: string = 'Updated';
+  sortKey: string = 'updated_at';
+  headers: IHeaders = {
+    updated_at: OrderBy.ASC,
+    created_at: OrderBy.ASC,
+  };
+
+  constructor(@Inject(forwardRef(() => TdManagementListComponent)) private _parent: TdManagementListComponent) {
```
emoralesb05: For a `design pattern` purpose, i think its better to just have the `sortBy` function here.

I get why we would want to reuse it, but we always want to push people to sort in the server, and not in the client.

kyleledbetter:
@richavyas this looks fantastic. As discussed let's keep the pattern to just mock & pattern, and update the spec in the issue https://github.com/Teradata/covalent/issues/260

So i'd recommend moving any logic the platform for now and keep this limited to design pattern in the docs app

richavyas:
- [x] Changed this to pattern & mock
- [x] replaced menu to `select` to load filter options

filter-sort-select

kyleledbetter:
Pushed updates:
manage-list-pattern

emoralesb05:
Pushed a couple of tweaks to leverage the usage of `md-select` and make the example more generic

PR #285, opened 2017-01-24
enhancement/docs/deployment (prod build clarification)

Archos67:
## Description

Clarified build process for production with the explanation below.

## Building for production

Adding the prod flag to ng build will set the build target and environment to production, which will optimize the build.

`ng build --prod`

## Ahead-of-time Compilation

When building for production, add the aot flag to enable the Ahead-Of-Time compiler where the browser will download a pre-compiled (from TypeScript) version of your application. Ahead-of-time compilation may require modifications to your code, but will enable faster rendering, fewer asynchronous requests, smaller Angular framework size, and better security.

`ng build --prod --aot`

### DIFF
--- a/src/app/components/docs/deployment/deployment.component.html
+++ b/src/app/components/docs/deployment/deployment.component.html
@@ -35,7 +35,6 @@
     of pushing to <code>gh-pages</code>, since user and organization pages require this.</p>
 
     <h3>Creating a build</h3>
-    <p></p>
     <p>The build artifacts will be stored in the dist/ directory.</p>
    <td-highlight lang="html">
      ng build
--- a/src/app/components/docs/deployment/deployment.component.html
+++ b/src/app/components/docs/deployment/deployment.component.html
@@ -35,9 +35,27 @@
     of pushing to <code>gh-pages</code>, since user and organization pages require this.</p>
 
     <h3>Creating a build</h3>
+    <p></p>
     <p>The build artifacts will be stored in the dist/ directory.</p>
    <td-highlight lang="html">
      ng build
+   </td-highlight>
+
+    <h3>Building for production</h3>
+    <p>Adding the <code>prod</code> flag to <code>ng build</code> will
+      set the build target and environment to production, which will optimize the build.</p>
+   <td-highlight lang="html">
+     ng build --prod
+   </td-highlight>
+
+    <h3>Ahead-of-time Compilation</h3>
+    <p>When building for production, add the <code>aot</code>
+      flag to enable the Ahead-Of-Time compiler where the browser will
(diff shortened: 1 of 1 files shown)


kyleledbetter:
thx for the PR @Archos67! reviewing now

DanjBethel:
Could double check but i think aot is default now with latest version of angular-cli?

@Archos67 @kyleledbetter

Archos67:
There is an [open issue](https://github.com/angular/angular-cli/issues/3368) related to the AOT compiler being enabled by default. I don't think that it is intended to be the default behavior.

I did a test myself on an older version and aot compilation was not the default behavior. Version details below.

angular-cli: 1.0.0-beta.22-1
node: 6.9.2
os: win32 x64

@DanjBethel

DanjBethel:
@Archos67 

Thanks for the clarification!

Will keep my eye on that

src/app/components/docs/deployment/deployment.component.html
```diff
@@ -39,6 +39,23 @@
...
+    <p>Adding the <code>prod</code> flag to <code>ng build</code> will
+      set the build target and environment to production, which will optimize the build.</p>
+   <td-highlight lang="html">
+     ng build --prod
+   </td-highlight>
+
+    <h3>Ahead-of-time Compilation</h3>
+    <p>When building for production, add the <code>aot</code>
+      flag to enable the Ahead-Of-Time compiler where the browser will
+      download a pre-compiled (from TypeScript) version of your application.
```
emoralesb05: This is not entirely accurate. What `AoT` does is pre-compile the `angular` modules before hand into something the `browser` can interpret.. instead of bundling the `angular/compiler` with the application and compiling everything on execution time in the browser (reason it takes a while to load on `JIT`).

`Typescript` is always compiled be it `AoT` or `JIT`.

kyleledbetter: @Archos67 can you change `enable the Ahead-Of-Time compiler where the browser will download a pre-compiled (from TypeScript) version of your application` to `pre-compile angular modules into something the browser can interpret instead of bundling the angular/compiler within the application and compiling everything in the browser` or similar?

Archos67:
Expanded on the AOT compiler explanation:

<p>When building for production, add the <code>aot</code>flag to enable the Ahead-Of-Time compiler. This will pre-compile Angular modules
      into native Javascript the browser can easily interpret instead of bundling the Angular compiler within
      the application and compiling everything in the browser. This will result in faster rendering,
      fewer asynchronous requests, smaller Angular framework size, and better security for your application.</p>

emoralesb05:
LGTM

PR #309, opened 2017-02-01
feature(release): Publish release to run as individual script

richavyas:
## Description

Added a new script specifically to publish the release instead of keeping it combined with finish release script following separation of concerns principle.

- [ ] `npm run lint` passes.


scripts/finish-publish
```diff
@@ -0,0 +1,11 @@
...
+set -e
+
+echo "Building master branch"
+npm run build
+
+echo "Publishing master branch"
+npm run publish
+
+echo "Return to develop branch"
+git checkout develop
```
emoralesb05: This should still be part of the `release` script since we move `explicitly` to `master` to merge stuff and we wanna go back to `develop` in it.

I think its better to keep this script clean from branch movements.

package.json
```diff
@@ -27,8 +27,9 @@
     "build": "bash scripts/build-release",
     "publish": "bash scripts/publish-release",
     "ghpages-deploy": "./node_modules/.bin/ng build --base-href /covalent/ --aot -prod && bash scripts/ghpages-deploy",
-    "start-release": "bash scripts/start-release",
-    "finish-release": "bash scripts/finish-release"
+    "start-release": "bash scripts/start-release", 
+    "finish-release": "bash scripts/finish-release",
+    "finish-publish": "bash scripts/finish-publish"
```
emoralesb05: maybe a good name would be `npm-publish`. Thoughts?

richavyas:
- [x] Better names to release scripts
- [x] `publish script` has only publish steps and not branch movements

emoralesb05:
- [x] Simplified it a bit more to just be `npm-publish` and leverage `package.json` for the commands.

PR #311, opened 2017-02-02
bugfix(build): cleanse the build by removing the `ts` source files.

emoralesb05:
## Description

There is a need to remove the source files (`ts`) from the release code since the `@angular-cli` picks them up instead of the `js` files which in term breaks `aot` with their new release.

We still needed to do this since those files shouldnt be there either way.

### What's included?

- Cleanse of build when running the `npm run build` script.

#### Test Steps

- [ ] `npm run build` in `develop` and see that all the `ts` files are in `deploy/platform/`
- [ ] `npm run build` in this branch and see that they are removed now.

- [ ] `npm run lint` passes.

### DIFF
--- a/scripts/build-release
+++ b/scripts/build-release
@@ -15,6 +15,12 @@ echo "Preparing files for AoT build"
 npm run aot
 echo "Compiled TS and generated *.metadata.json files..."
 
+# Clean source .ts files
+cd deploy/
+find . -name "*.ts" ! -name "*.d.ts" -type f -delete
+cd ..
+echo "Remove source .ts files so they arent published"
+
 # Inline the css and html into the component files.
 gulp inline-resource-files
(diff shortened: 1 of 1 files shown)


scripts/build-release
```diff
@@ -18,6 +18,12 @@ echo "Preparing files for AoT build"
 npm run aot
 echo "Compiled TS and generated *.metadata.json files..."
 
+# Clean source .ts files
+cd deploy/
+find . -name "*.ts" ! -name "*.d.ts" -type f -delete
```
richavyas: is it supposed to delete `d.ts` files too because I still see those in `deploy` folder?

emoralesb05: Nope, `d.ts` are needed since those are the type definition files for typescript.

PR #321, opened 2017-02-04
performance(modules): Only import modules that are required. (closes #287)

### LINKED ISSUE #287: Modules optimization
#### Feature Request

Decrease imported components from Covalent modules by importing only material modules that are necessary but not the whole `MaterialModule.forRoot()`

Here is an example. I do not import/use material in my app at all. I import just Covalent steps module and the result is 64KB for Covalent steps and 745KB for material in my bundle.
I understand that steps module uses some material components but I'm pretty sure that input, slider, dialog and many more material components is not used in Covalent steps component.
steps

Import only necessary material modules (like MdCheckboxModule or MdButtonModule and so on) in Covalent component's modules but not  the whole module `MaterialModule.forRoot()`.

Using Covalent component bloats my bundle size with >700K because of including all material components in it.

All

emoralesb05:
## Description

There is an issue `tree shaking` because of how Typescript + Webpack compile and bundle things.

https://github.com/angular/angular-cli/issues/2901
https://github.com/webpack/webpack/issues/2899

We do not want to depend on tree shaking alone (if they fix it great, if not doesnt matter), so we need to *explicitly import* the modules that are required and nothing else. https://github.com/Teradata/covalent/issues/287

For now we will just assume that if you import `CovalentCoreModule` you want everything, but if you import a `submodule` then we will only use the required ones.

### What's included?

- Removal of generic `MaterialModule.forRoot()` imports and only adding required ones per `module`

e.g.

Search module only requires the following modules:

```typescript
    FormsModule,
    CommonModule,
    MdInputModule.forRoot(),
    MdIconModule.forRoot(),
    MdButtonModule.forRoot(),
```

#### Test Steps

- [x] `ng serve`
- [x] See the docs still work with only the required modules.

- [x] `npm run lint` passes.


src/platform/core/notifications/notifications.module.ts
```diff
@@ -2,7 +2,7 @@ import { Type } from '@angular/core';
 import { NgModule, ModuleWithProviders } from '@angular/core';
 
 import { CommonModule } from '@angular/common';
-import { MaterialModule } from '@angular/material';
+import {  } from '@angular/material';
```
richavyas: What's the intent behind leaving this empty?

emoralesb05: To see if people would notice haha jk, my bad

src/platform/dynamic-forms/dynamic-elements/dynamic-select/dynamic-select.component.html
```diff
@@ -2,7 +2,7 @@
   <md-select [(ngModel)]="value"
              [placeholder]="label"
              [required]="required"
-             flex>
+             flex="100">
```
richavyas: Probably I am not aware, how `flex` is different from `flex="100"`

PR #328, opened 2017-02-07
feature(ngx-translate): added documentation on how to use `ngx-translate`

emoralesb05:
## Description

There is a need for `i18n` and `l10n` in all our products, so from all the options out there, `ngx-translate` seemed the easiest one to use + filled all our needs.

### What's included?

- Utility (internal) methods for `ngx-translate` usage.
- Demo for `ngx-translate` usage with `[ngPlural]` http://cldr.unicode.org/index/cldr-spec/plural-rules
- Demo for `ngx-translate` usage with `LOCALE_ID` and angular pipes.
- Demo for `ngx-translate` usage with `tdDialogService`

#### Test Steps

- [x] `ng serve`
- [x] Go to http://localhost:4200/#/components/ngx-translate
- [x] See demo and play around with it.

- [x] `npm run lint` passes.


src/app/assets/i18n/en.json
```diff
@@ -0,0 +1,28 @@
+{
+  "DEMO_TAB": "Demo",
+  "CODE_TAB": "Code",
+  "DEMO_ONE": {
+    "TITLE": "Examples with ngx-translate",
+    "SUBTITLE": "i18n with ngx-translate + [ngPlural] from Angular",
+    "SELECT_TOOLTIP": "When this value changes, [ngPlural] reads it to select the correct message",
+    "COUNT_UNSELECTED": "The value is undefined",
```
richavyas: I would rather not show this message. When I saw this message on page load, it appeared as if it's shown because of some error or wrong parameter. We can directly show `There are no selected values` instead.

richavyas:
Observed some load wonkiness. It's loading intermittently:

translate-component-wonkiness

emoralesb05:
Not sure i understand the `gif`, what was the issue?

richavyas:
The gif shows that all other modules are clickable and working fine but nothing happens after ngx-translate component is clicked. It hangs.

emoralesb05:
Anything in the `console.log`? cant seem to replicate it

src/app/assets/i18n/en.json
emoralesb05: Good point, ill just remove that part.

emoralesb05:
- [x] Added a more basic example
- [x] Fixed interpolation issues on examples.

Issue #324, opened 2017-02-05
[Enhancement] Test (and modify) all our components to make them support or handle `i18n` and `l10n`.


emoralesb05:
#### Feature Request

We need to make sure all our components either handle `i18n` and `l10n`, or atleast are flexible enough so the text could be replaced.

Covalent Modules:

- [ ] Common (Pipes)
  - [x] Digits
  - [ ] Bytes
  - [ ] TimeAgo
  - [ ] TimeDifference
- [x] Chips
- [x] Data Table
- [x] Dialogs
- [x] Expansion Panel
- [x] File
- [x] Json Formatter
- [x] Layout
- [x] Menu
- [x] Notifications
- [x] Paging
- [x] Search
- [x] Steps
- [x] Dynamic Forms
- [x] Highlight
- [x] Markdown

forrest-lyman:
One thing (I work a lot outside the US) that I really appreciate is either opting in to localized text or having a message letting me know that i am seeing translated content with a prominent link to the default language.

crain:
Would this only support i18n the Angular2 native way or would it be possible to override labels via @Input parameters to use frameworks like https://github.com/ngx-translate/core ?

emoralesb05:
@forrest-lyman that is more of an issue on implementation on the product side, but its a nice feature we can implement in our internal products.

@crain For the most part we want every piece of text to be provided on consumption through `inputs`/`templates` to allow that flexibility. Maybe only a few might have hardcoded english strings that can be overridden by those inputs.

We are probably gonna start using `ngx-translate` ourselves, so we want to work well with it.

quanterion:
@emoralesb05 Why are you going to use ngx-translate? I can see one showstopper to use native Angular i18n - that is translating text inside source files. But @ocombe, the creator of ngx-translate is now a core contributor in Angular and going to solve this issue soon https://github.com/angular/angular/issues/11405

PS: I do think native angular i18n is more efficient since it produce no runtime penalty and all hard work is done is compilation stage that can be significant for big enterprise apps.

ocombe:
I am working on this right now, there is currently a blocker because I need to wait for the implementation of TypeScript transformers into tsc_wrapper for Angular (the team is working on this), but if everything goes well, this should land around 4.2 (maybe 4.3). This is a rough estimate since the blocker is not up to me :-)

emoralesb05:
[...]

PR #333, opened 2017-02-10
chore(): upgrade @ngx-translate to 6.0.0-beta.1

emoralesb05:
## Description

Upgrade `ng2-translate` to `@ngx-translate/core` and also include `@ngx-translate/http-loader`
https://github.com/ngx-translate/http-loader#ngx-translatehttp-loader--

#### Test Steps

- [x] `npm run reinstall`
- [x] See new resources card in `ngx-translate` docs.

- [x] `npm run lint` passes.


kyleledbetter:
Pushed a couple tweaks adding the translate buttons on each card w/ diff color:

image
image

lincze:
does that mean that e.g. in date pipe '6 months ago"-like messages can also be translated?

emoralesb05:
That one is a different issue that we are gonna tackle, but soon it will follow the `LOCALE_ID` like any other pipe.

We have this issue open to track that. https://github.com/Teradata/covalent/issues/324

lincze:
thanks, I would have to find it

PR #334, opened 2017-02-10
feature(dependencies): add @angular/flex-layout (closes #239)

### LINKED ISSUE #239: [Feature] Integrate @angular/flex-layout
#### Feature Request

I've integrated flex-layout successfully in one of my work apps already and have been impressed with how much it simplified the layout portion of adding new features. Seems like it would be a good fit here assuming you are interested.

Also worth noting that the `MatchMediaObservable` has all sorts of cool possibilities for UX.

In my opinion `Responsive layout is where free time goes to die` so anything that can simplify that for developers is a good thing :)

The proposal would be to add the as yet, unreleased @angular/flex-layout v2.0.0-beta.2 as it has two needed bug fixes and introduces a custom break points feature.

##### Other information
I'm happy to do the work assuming all the stake holders want to go this route. The end result being to get back to exactly where this is now just using flex-layout to get there.

emoralesb05:
## Description

Including `@angular/flex-layout` as part of `CovalentCoreModule` and started initial migration to leverage it. https://github.com/Teradata/covalent/issues/239

This is meant to show both `flex` libs can co-exist in `covalent` as we migrate into `@angular/flex-layout`.

For the time being, we will be fixed to a specific version since its a beta release. (beta.5)

### What's included?

- chore(): add `@angular/flex-layout` as as a `@covalent/core` dependency
- feature(paging): leverage flex-layout in paging module.
- feature(search): leverage flex-layout in search module.

#### Test Steps

- [x] `npm reinstall`
- [x] `ng serve`
- [x] Go to http://localhost:4200/#/components/search and see it work fine
- [x] Go to http://localhost:4200/#/components/paging and see it work fine
- [x] Try using the `@angular/flex-layout` anywhere in docs.

- [x] `npm run lint` passes.


kyleledbetter:
Code & UX LGTM 👍 

I've got questions more around the performance & cascading implications of flex-layout using inline styles instead of CSS stylesheets, that are mainly just unknowns for me, since we use flex layout not only for layouts but for every single UI element. That's a ton of extra inline HTML styles...

Obviously I can see all the benefits of this being in JavaScript with all the power of Angular2 with media queries, change detection, etc, which is why I'm excited about it

I'll do more research from their wiki and more testing on my own.

d3viant0ne:
@kyleledbetter - I started using `fxFlex` at work back when it was still private and you are correct there is definitely a balance to be found between inline & CSS.

What I ended up settling on was using `fxFlex` for container components layout & anything where benefits `ObservableMedia` are too good to pass up. All the low level styles for components I ended up moving back to .css, not so much because of a performance issue but simply because of the bloat in my HTML styles.

emoralesb05:
@d3viant0ne Yeah, it feels like we might have a need for both things to coexist for now.

I still think its greatly needed (it has awesome features) but we might stick to ours for internal usage in our components and use `@angular/flex-box` for more granular stuff maybe within applications and docs.

kyleledbetter:
@emoralesb05 i'm cool with merging in for use if we revert search & paging to use the css classes, that way we can do more testing downstream in products

emoralesb05:
@kyleledbetter yep, will revert those so we can just merge it

PR #344, opened 2017-02-15
Typo Fix: Anguer -> Angular

nakulgulati:
## Description
Typo fix on home page FAQ section


emoralesb05:
yikes. thanks for that!

PR #348, opened 2017-02-17
fix(rollup): add missing dep to options.globals

d3viant0ne:
## Description

Flex-Layout was added as a dependency but was omitted from the rollup script.

### What's included?

- Adds flex-layout prop to rollup.js `options.globals` .

#### Test Steps

Executing  `build-release` will currently emit a warning when executing rollup.

```
Treating '@angular/flex-layout' as external dependency
No name was provided for external module '@angular/flex-layout' in options.globals – guessing '_angular_flexLayout'
Treating '@angular/flex-layout' as external dependency
No name was provided for external module '@angular/flex-layout' in options.globals – guessing '_angular_flexLayout'
```


emoralesb05:
Oh snap! haha thanks for that.

PR #305, opened 2017-02-01
feature(material): Upgrade to latest material. (closes #346)

emoralesb05:
## Description

Upgrade to latest `@angular/material@2.0.0-beta.2`. https://github.com/angular/material2/blob/master/CHANGELOG.md#200-beta2-flannel-papaya-2017-02-15

### What's included?

Material Changes:
- Breaking change from `material` where `md-input` has been removed as component and directive in favor of the `mdInput` directive selector.
- Leverage new input of `md-menu` `[overlapTrigger]="false"` in notifications demo so the menu doesnt overlap the `icon`
- Now we can use the `(change)` event from `md-select`.

Covalent Changes/Fixes because of `material`:
- **internal(efficiency):** No more need for `refresh` work around for `tooltips` used under `OnPush` https://github.com/angular/material2/commit/632b964b0dd1cf15349265ec118a6844e6cec9d0
- **bugfix(layout):** Leverage `disableClose` input from sidenav to prevent `ESC` closing the sidenav on `side` mode. https://github.com/angular/material2/commit/52ade97cfc91d22b9d5b0c77c7bd09581816f5a0
- **bugfix(dynamic-forms):** `number` type input works again.
- **feature(data-table):** Leverage the `md-pseudo-checkbox` to render the selectable rows more efficiently. https://github.com/angular/material2/commit/3b6cab01c7ff44c999ffcdc99d2ec9a2fb63e96a

#### Test Steps

- `npm run reinstall`
- `ng serve`
- Play around with docs

#### General Tests for Every PR

- [x] `ng serve --aot` still works.
- [x] `npm run lint` passes.
- [x] `npm test` passes and code coverage is not lower.
- [x] `npm run build` still works.

NOTE: Needs heavy testing on all fronts since the `.mat` prefix change affected how certain rules are applied.

PR #360, opened 2017-02-22
Update directives.component.html

MingyuJeon:
Updated md-input to mdInput


emoralesb05:
Thanks for the contribution, but we are already taking care of this in https://github.com/Teradata/covalent/pull/305.

PR #377, opened 2017-02-24
fix(): remove @Injectable decorator from HttpInterceptorService. (closes #340)

### LINKED ISSUE #340: [Module: HTTP] - AOT compile fails with angular@next
This repository's issues are reserved for feature requests and bug reports.

OS: macOS 10.12.3
NodeJS: Both 6.9.5 (all ng versions) && 7.5.0 (covalent on ng 2.4.7 & 4.0.0-beta.7 only)
Browser: 58.0.3012.0 canary (64-bit)

#### Bug Report

- Update to `@angular/* 4.0.0-beta.7`
- Execute `./node_modules/.bin/ng build --base-href /covalent/ --aot -prod`

AOT Compile fails in `http-interceptor.service.ts` specifically `config.interceptors` isn't resolving.

This is more of a heads up than a bug as there is no real expectation of Angular 4.x support at the moment though imo worth logging as the lib could be @next ready with one fix.

Outside of the one AOT issue, Covalent actually works just fine on 4.x. Actually, it's rendering a bit faster on beta.7 than it does on 2.4.7. as well as "feeling" a bit smoother. Unit tests execute without issue, manually verified consistent behaviors for Covalent running on 2.4.7 & 4.0.0-beta.7 within the confines of the demo application.

Tested with `4.0.0-beta.2` through `4.0.0-beta.7` as well as the latest nightly with the same result.

##### Other information
(e.g. detailed explanation, stacktraces, related issues, suggestions how to fix)

_AOT Output_
```
ERROR in Can't resolve all parameters for HttpInterceptorService in /Users/d3viant0ne/Projects/oss/teradata/covalent/src/platform/http/interceptors/http-interceptor.service.ts: ([object Object], [object Object], [object Object], ?).

ERROR in ./src/main.ts
Module not found: Error: Can't resolve './$$_gendir/app/app.module.ngfactory' in '/Users/d3viant0ne/Projects/oss/teradata/covalent/src'
 @ ./src/main.ts 4:0-74
 @ multi ./src/main.ts
```

emoralesb05:
## Description

https://github.com/Teradata/covalent/issues/340
we can remove this decorator since this service isnt intended to be injected alone, it needs to be provided via factory. (e.g. Router)

### What's included?

- Removal @Injectable decorator from HttpInterceptorService so we can support @angular@4.X

#### Test Steps

- [ ] `ng serve --aot` in current version (see it working)
- [ ] point to `@angular/*/@4.0.0.beta.8` or up in the `package.json`
- [ ] `npm run reinstall`
- [ ] `ng serve --aot`

- [ ] `npm run lint` passes.

### DIFF
--- a/src/platform/http/interceptors/http-interceptor.service.ts
+++ b/src/platform/http/interceptors/http-interceptor.service.ts
@@ -13,7 +13,6 @@ export interface IHttpInterceptorConfig {
   paths: string[];
 }
 
-@Injectable()
 export class HttpInterceptorService {
 
   private _requestInterceptors: IHttpInterceptorMapping[] = [];


d3viant0ne:
@emoralesb05 - Kind of feel like the @next grim reaper here but you may want consider just incorporating this with all the other changes that will be required for angular 4.0.0-rc.0 + whatever they do with material to contend with the `template` => `ng-template` headache

That said, the `@Injectable` change works just fine thus far. No apparent issues using the `HttpInterceptorService` over the weekend.

Issue #323, opened 2017-02-05
[Enhancement] Test (or Add) the  `rtl` support to all modules and docs.


emoralesb05:
#### Feature Request

We need to make sure every component (and the docs) work fine in `rlt` and `ltr`.

Covalent Modules:

- [ ] Common
  - [ ] pad styles
  - [ ] pull styles
  - [ ] push styles
- [x] Chips
- [x] Data Table
- [x] Dialogs
- [x] Expansion Panel
- [x] File
- [x] Json Formatter
- [x] Layout
- [x] Menu
- [x] Notifications
- [x] Paging
- [x] Search
- [x] Steps
- [x] Dynamic Forms
- [ ] Highlight
- [ ] Markdown

Docs:

- [ ] Documentation
- [ ] Style Guide
- [ ] Layouts
- [ ] Components

akhedrane:
hi, i would thank you for this really great work,
I'am a native RTL user so i made a pull request with some change to support RTL direction at #406

emoralesb05:
@akhedrane 💥  awesome! thanks for all the help. 

Will find time to review/test over the week. 😄

emoralesb05:
Closing this since we support RTL in most of our components now.

PR #413, opened 2017-03-09
feat(ltr): add fallback styles for `ltr` + additional support.

emoralesb05:
## Description

Part of https://github.com/Teradata/covalent/issues/323
Instead of assuming there will be a `dir="ltr"` or `dir="rtl"`, there is a need to set defaults and then overrides when its `dir="rtl"`, since there are use cases where devs wont use `dir` at all.

### What's included?

- feat(rtl): added default rules for `ltr` support with overriding `rlt` rules.
- chore(rtl): polish usage of `rtl`
- bugfix(rtl): fixed cases where `rtl` wasnt being applied.
- feat(file-upload): added support for `rtl`.
- feat(json-formatter): added somewhat support for `rtl` (there is still issues around the strings.
- feat(search): added support for `rtl`.
- feat(paging): added suport for `rtl`.
- bugfix(notifications): fixed `rtl` support.
- bugfix(data-table): fixed `rtl` support.

#### Test Steps

- [ ] `ng serve`
- [ ] Go to http://localhost:4200/#/components
- [ ] Test components on both `ltr` and `rtl` 

- [ ] `npm run lint` passes.

NOTE: we still need to figure out how to handle strings like `Array [n]` , etc etc in `rtl`

Also need to figure out how are we gonna handle `rtl` for utility classes like `pad`, `push` and `pull`


akhedrane:
for utility classes, i think naming like `pad-start` & `pad-end` is more semantic.

emoralesb05:
@akhedrane yeah, i was actually thinking something like that or `pad-before` and `pad-after` to be inline with `material`

kyleledbetter:
We can leave the current left/right/top/bottom for explicit versions (they do this also in some places) and add additional before/after/above/below that are dynamic

PR #458, opened 2017-03-28
Rename utilities to mark as partial and prevent import conflicts

DaSchTour:
Preventing the following error when trying to build platform.scss from node_modules inside another application.

```
Message:
    node_modules/@covalent/core/common/platform.scss
Error: It's not clear which file to import for '@import "styles/utilities/utilities"'.
       Candidates:
         styles/utilities/utilities.scss
         styles/utilities/utilities.css
       Please delete or rename all but one of these files.
        on line 28 of node_modules/@covalent/core/common/platform.scss
>> @import "styles/utilities/utilities";
```


kyleledbetter:
Good catch, thanks @DaSchTour!

Issue #461, opened 2017-03-29
sort for tables not working with date strings


pascalwhoop:
#### Bug Report

calling sortData on TdDataTableService with sorting field being a date doesn't work as expected

#### What is the expected behavior?
sorting date should sort the dates. But the line
`if (!Number.isNaN(Number.parseFloat(compA)) && !Number.isNaN(Number.parseFloat(compB)))`

causes Problems. This is beacuse `Number.parseFloat("2018-05-14T00:00:00.000Z")` returns a number. 2018

#### What is the motivation / use case for changing the behavior?
Dates should be sorted properly

akomarovsky:
was it fixed? I see that in version 1.0.0-beta.3-1 it is still an issue

pascalwhoop:
No it wasn't fixed but there is no clean way of sorting this without being really careful what you do. I suggested including moment.js but the repo admins don't agree that's a good idea yet so I don't have a good idea at the moment. Doing some sketchy Regex on strings is not a smart move.

emoralesb05:
We are trying to avoid adding external dependencies to reduce the bundle size. Unless the dependency is used heavily in the project, its not really worth it. We want to be able to fix this without it if possible.

That being said, you can always pull `moment.js` on your project and use it.

michael-lang:
Can you just do a custom sort?  See code snippet in #521

pascalwhoop:
Yes of course it's possible to get my date strings sorted manually. And although we shouldn't expect covalent to figure out for us what the string means, a date string is so common it would be nice. The real problem however is that it currently(is it still current?) interprets a date string as a number which becomes the year of the date. That is not correct. I can't look into the code base ATM, I'm on mobile. 

I think best would be to rethink the interpretation of what is a number and leave the date handling out until either moment.js is included or someone finds a solid well proven way of determining strings being ISO date strings or not.

ankita3076:
Hi, 
I am able to sort data manually, but the problem is when I am clicking on pagination, all the sorting has gone. Can anyone help me in that.

Vasanthvivi:
> We are trying to avoid adding external dependencies to reduce the bundle size. Unless the dependency is used heavily in the project, its not really worth it. We want to be able to fix this without it if possible.
> 
> That being said, you can always pull `moment.js` on your project and use it.

any other thing other than moment?

PR #462, opened 2017-03-29
fix(dataTableService): Sorting simplified and fixed for date type

pascalwhoop:
Let's just leave javascript to handle the sorting. If its a number, it will sort correctly, same goes for strings and dates/strings of dates.

related to issue #461

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "covalent",
-  "version": "1.0.0-beta.2-1",
+  "version": "1.0.0-beta.3",
   "private": true,
   "description": "Teradata UI Platform built on Angular Material",
   "keywords": [
--- a/src/platform/core/package.json
+++ b/src/platform/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@covalent/core",
-  "version": "1.0.0-beta.2-1",
+  "version": "1.0.0-beta.3",
   "description": "Teradata UI Platform built on Angular Material",
   "main": "./core.umd.js",
   "module": "./index.js",
--- a/src/platform/dynamic-forms/package.json
+++ b/src/platform/dynamic-forms/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@covalent/dynamic-forms",
-  "version": "1.0.0-beta.2-1",
+  "version": "1.0.0-beta.3",
   "description": "Teradata UI Platform Dynamic Forms Module",
   "main": "./dynamic-forms.umd.js",
   "module": "./index.js",
@@ -36,6 +36,6 @@
     "Jeremy Smartt <jeremy.smartt@teradata.com>"
   ],
   "dependencies": {
-    "@covalent/core": "1.0.0-beta.2-1"
+    "@covalent/core": "1.0.0-beta.3"
   }
 }
--- a/src/platform/highlight/package.json
+++ b/src/platform/highlight/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@covalent/highlight",
-  "version": "1.0.0-beta.2-1",
+  "version": "1.0.0-beta.3",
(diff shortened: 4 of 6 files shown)


emoralesb05:
The reason we had it that way is because there are times we want to sort numbers that are strings. 

Not everything will be in its correct type.

pascalwhoop:
A number that is a string will still sort correctly though as "1" is still
smaller than "11" or "19" isn't it?

emoralesb05:
In the case of "11" and "19" yes, but if you compare "100" with "20", turns out that "100" will be taken as lower than "20" because `string`s are compared character by character left to right.

This is true in all the languages (or at least the ones i know)

pascalwhoop:
Okay, true. Mhm maybe an additional attempt of parsing the string as a date
before parsing it as a number would be a good approach? If you'd think that
is a better idea, I'll update the PR, adding a commit that does
1. check if date --> date compare
2. check if number --> number compare
3. assume string --> string compare

emoralesb05:
Yep, that would be awesome! 😄

pascalwhoop:
How would you think about including moment.js in the project? I have reservations about my proposed solution as it seems to be a more complex field than I originally anticipated
https://stackoverflow.com/questions/7445328/check-if-a-string-is-a-date-value

They argue that its not as intuitive as one would think (e.g. just checking if `Date.parse('somestring')` returns "Invalid Date" or not. So I think this needs to be done properly rather than me just giving it a quick fix. I could include moment in the build and use the library for that if you agree with adding this dependency.

pascalwhoop:
The alternative (IMHO somewhat dirty) solution would be something in the lines of 

```
if(isNaN(compA) && isNaN(compB) && isValidDate(compA) && isValidDate(compB){
    // do date comparison
}else{
    // same code as before
}

**Comment by emoralesb05** [comment]
Well, the reason we abstracted the sorting function from the data-table was so people can implement their own. Be it server side or any other type of sorting algorithm + we didnt want to include `moment.js` as part of our library for the time being if possible. 

I agree that the issue is not simple to fix, so it might require a bit more brainstorming before continuing.

**Comment by RemcovandenBerg** [comment]
Workaround.. 
```
 this._dataTableService.sortData = 
      (data: any[], sortBy: string, sortOrder?: TdDataTableSortingOrder): any[] => {

        let looksLikeAnISODate = (compA: any): boolean => {
          if (!compA) return false;
          if (compA.length < 19) return false;
          return (compA[4] == '-') && (compA[7] == '-') && (compA[10] == 'T');
        }

        let determinedIsoSort = false;
        let isoSort = false;
        let getIsIsoDateColumnSortable = (compA, compB: any) : boolean =>{
            if (determinedIsoSort) return isoSort;
            isoSort = looksLikeAnISODate(compA) || looksLikeAnISODate(compB);
            determinedIsoSort = !!compA || !!compB;
            return isoSort;
        }        

        if (sortOrder === void 0) { sortOrder = TdDataTableSortingOrder.Ascending; }
        if (sortBy) {
            data = Array.from(data);
            data.sort(function (a, b) {
                var compA = a[sortBy];
                var compB = b[sortBy];
                
                var direction = 0;
                let sortAsNumber =  !Number.isNaN(Number.parseFloat(compA)) && !Number.isNaN(Number.parseFloat(compB));
                let isIsoDateColumn = getIsIsoDateColumnSortable(compA, compB);
                if (sortAsNumber && !isIsoDateColumn) {
                    direction = Number.parseFloat(compA) - Number.parseFloat(compB);
                }
                else {
                  if (!compA) compA = "";
                  if (!compB) compB = "";

                    if (compA < compB) {
                        direction = -1;
                    }
                    else if (compA > compB) {
                        direction = 1;
                    }
                }
                return direction * (sortOrder === TdDataTableSortingOrder.Descending ? -1 : 1);
            });
        }
        return data;
    };
```

PR #464, opened 2017-03-29
feat(dependencies): upgrade @angular/material to `beta.3`.

emoralesb05:
## Description

Upgrade covalent to use and leverage the latest material.

### What's included?

- fix(theming): update theming to point to the correct path
- fix(layout): properly style `md-sidenav` without the `cdk-focus-trap` element.
- fix(chips): replace [dividerColor] with [color] in md-input-container
- perf(steps): remove MdListModule dependency from CovalentStepsModule
- perf(expansion-panel): remove MdListModule dependency from CovalentExpansionPanelModule
- fix(validations): remove input invalid style
- chore(): deprecate CovalentCoreModule

#### Test Steps

- [ ] `ng serve`
- [ ] Test the app

- [ ] `npm run lint` passes.


d3viant0ne:
The eagle has landed so to speak :) And evidently we are back to importing components on an "as needed" basis.

asaph26:
It would be great if there are independent imports of the material modules rather than MaterialModule as a whole given that the MaterialModule is deprecated

emoralesb05:
@asaph26 we probably will deprecate `CovalentCoreModule` too, so we push people to import only what they need.

src/app/components/components/material-components/material-components.component.html
```diff
@@ -1329,27 +1329,27 @@ <h3 md-line>{ { dog.name } }</h3>
   <md-card-title>Slider</md-card-title>
   <md-divider></md-divider>
   <md-card-content>
-    <md-slider color="accent" min="1" max="5" thumbLabel tickInterval="1"></md-slider>
-    <md-slider step="5" thumbLabel tickInterval="5"></md-slider>
-    <md-slider thumbLabel tick-interval="auto"></md-slider>
+    <md-slider min="1" max="5" thumbLabel tickInterval="1" value="2"></md-slider>
```
jeremysmartt: why do these have values now?

emoralesb05: when the value is undefined, it shows black.

We wanted to show the accent colors always in the demos.

PR #526, opened 2017-04-27
Add sortOrder property to ITdDataTableColumn interface

samaxes:
Useful when using [Data Table (Atomic) Components](https://teradata.github.io/covalent/#/components/data-table).

### DIFF
--- a/src/platform/core/data-table/data-table.component.ts
+++ b/src/platform/core/data-table/data-table.component.ts
@@ -28,6 +28,7 @@ export interface ITdDataTableColumn {
   format?: (value: any) => any;
   nested?: boolean;
   sortable?: boolean;
+  sortOrder?: TdDataTableSortingOrder;
 };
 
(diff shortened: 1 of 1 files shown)


emoralesb05:
You can just extend it and create your own custom one.

The problem with adding this is that in the normal case of the `data-table` component, it does have real use and it can be misleading as having it.

Having said this, we will probably need this soon when we make the data-table multi-sort compatible.

samaxes:
@emoralesb05 yep, that's what I'm doing while it's not supported directly by data-table.

Is there any issue I can follow about the data-table multi-sort implementation status? And are there any docs available so I can have a look of how it works?

emoralesb05:
Not yet, but we will add the issue soon since we need to develop it for our internal products too.

samaxes:
Thanks.

emoralesb05:
Closing this since we are in the design process on how to do multi sort.

PR #547, opened 2017-05-03
feat(chips): ability to disable chip addition (input). (closes #500)

### LINKED ISSUE #500: Feature Request: Turn off autocomplete for chips
#### Feature Request: Turn off autocomplete for chips
Currently, chips are only available with autocomplete, while this a great feature sometimes it can be overkill for curtain use cases.

#### Example Template HTML
```
<td-chips placeholder="Enter string"
[autoComplete]="autoComplete"
[items]="items"
[(ngModel)]="model"
[readOnly]="readOnly"
(add)="add($event)"
(remove)="remove($event)"
requireMatch>
</td-chips>
```
where the autoComplete is type boolean to turn on or off the auto complete feature. [autoComplete]="autoComplete"

[chips.component.html](https://github.com/Teradata/covalent/blob/develop/src/platform/core/chips/chips.component.html)
```
<div *ngIf="autoComplete">    
<md-input-container floatPlaceholder="never"
                        [style.width.px]="readOnly ? 0 : null"
                        [color]="matches ? 'primary' : 'warn'">
      <input mdInput
              flex="100" 
              #input
              [mdAutocomplete]="autocomplete"
              [formControl]="inputControl"
              [placeholder]="readOnly? '' : placeholder"
              (keydown)="_inputKeydown($event)"
              (keyup.enter)="addChip(input.value)"
              (focus)="handleFocus()"
              (blur)="handleBlur()">
    </md-input-container>
    <md-autocomplete #autocomplete="mdAutocomplete">
      <ng-template let-item ngFor [ngForOf]="filteredItems | async">
        <md-option (click)="addChip(input.value)" [value]="item">{{item}}</md-option>
      </ng-template>
    </md-autocomplete>
    </div>

```
[chips.component.ts ](https://github.com/Teradata/covalent/blob/develop/src/platform/core/chips/chips.component.ts)
```
  /**
   * autoComplete?: boolean
   * disables autocomplete on chips.
   * Defaults to 'true' or autocomplete enabled.
   */
  @Input('autoComplete') autoComplete: boolean = true;
```

##### Example
chips-example

JoshSchoen:
## Description
Feature Request: Turn off autocomplete for chips #500

### What's included?
* modified: src/platform/core/chips/chips.component.ts
* modified: src/platform/core/chips/chips.component.html
* modified: src/app/components/components/chips/chips.component.html
* modified: src/app/components/components/chips/chips.component.ts

#### Test Steps

- [ ] Checkout Branch
- [ ] ng serve
- [ ] Navigate to http://localhost:4200/#/components/chips
- [ ] Click on Toggle Autocomplete Button
- [ ] Make sure Autocomplete turns on and off

#### Manual Test

- [ ] edit src/app/components/components/chips/chips.component.ts
- [ ] Change autocomplete to false
- [ ] Save
- [ ] Navigate to http://localhost:4200/#/components/chips
- [ ] Make sure autocomplete is disabled under Chips & Autocomplete
- [ ] edit src/app/components/components/chips/chips.component.html
- [ ] Remove [autoComplete]="autoComplete"
- [ ] Save
- [ ] Navigate to http://localhost:4200/#/components/chips
- [ ] Make sure autocomplete is enabled under Chips & Autocomplete

- [ ] `npm run lint` passes.

auto-complete-chips


src/platform/core/chips/chips.component.ts
```diff
@@ -96,6 +99,12 @@ export class TdChipsComponent implements ControlValueAccessor, DoCheck, OnInit {
     return this._readOnly;
   }
 
+    /**
+   * autoComplete?: boolean
+   * Disables autocomplete. If it doesn't exist autocomplete defaults to true.
+   */
+  @Input('autoComplete') autoComplete: boolean = true;
```
emoralesb05: So calling the input `autocomplete` can make it misleading about what it actually does since the `autocomplete` part is technically an add-on to the `mdInput`.

I think we should call the property `allowAdd` or something like this since later on we can possibly add later on an `allowRemove` which can enable/disable the `x` buttons from the chips.

emoralesb05:
Also, can you fix the lint issues?

src/platform/core/chips/chips.component.ts
JoshSchoen: Will do I will make the change to allowAdd and the one bellow.Thanks!

JoshSchoen:
@emoralesb05 updated autoComplete to allowAdd and fixed the lint issue. Does the dependency-ci check take a while or is that a manual thing?

emoralesb05:
Not sure why it hangs when its a PR from a fork 🤔 .. but dont worry about it haha 😄 

Awesome dude! Gonna test the PR tonight 🎉

src/platform/core/chips/chips.component.html
```diff
@@ -9,6 +9,7 @@
         </md-icon>
       </md-basic-chip>
     </ng-template>
+    <div *ngIf="allowAdd">
     <md-input-container floatPlaceholder="never"
```
emoralesb05: Can you add indentation for the content of the `<div>`? 😄

emoralesb05:
There is a need to check for `allowAdd` in the `focus()` since it fails when focusing it when `allowAdd` is false.

e.g.

```
focus(): void {
    if (this.allowAdd) {
      this._inputChild.focus();
    }
}
```

src/platform/core/chips/chips.component.html
```diff
@@ -23,11 +24,13 @@
               (focus)="handleFocus()"
               (blur)="handleBlur()">
     </md-input-container>
+
     <md-autocomplete #autocomplete="mdAutocomplete">
```
emoralesb05: There is a weird bug when you toggle the `allowAdd` to false and then back to true, once you focus the `input` the autocomplete doesnt appear until you type on it.

The fix is to put the `md-autocomplete` out of the `<div *ngIf="allowAdd">` so its added as soon as the `md-input` is rendered without affecting the autocomplete rendering.

JoshSchoen:
Sounds good I'll make the changes and push again. Thanks!

src/platform/core/chips/chips.component.html
JoshSchoen: When I move the `md-autocomplete` outside `<div *ngIf="allowAdd">` it breaks the ability add chips, the input.value is undefined. One way to get around this to visually with CSS hide it rather than removing it from the DOM with *ngIf. Let me know if this is acceptable otherwise I can look into an alternative.

JoshSchoen: The alternative approach would be to change `(click)="addChip(input.value)` to (click)="addChip(item). I might be wrong but for the click event wouldn't this be a little more concise to match the available list if we are only allowing the items in the list array anyway? Of course the input is a different thing but I don't see any errors and the focus event works as excepted when toggle from false back to true.

emoralesb05: Yeah, I think it should use `addChip(item)` either way to be more consistent.

JoshSchoen: cool thanks!

kyleledbetter:
small thing, but for the API should `allowAdd` be named something more like `addChips`?  or since the real goal is disabling the input should it be more like `disableInput`?

kyleledbetter:
Also, if the input is disabled, the `.mat-input-underline` should be hidden, since underline is to imply that there's an input text field

kyleledbetter:
this would have an `*ngIf` conditional on it https://github.com/Teradata/covalent/blob/1994564738b31bd98152c984eda33a10c2f0a993/src/platform/core/chips/chips.component.html#L32

emoralesb05:
@kyleledbetter the reason i wanted to call it `allowAdd` is because after this there will be an addition of `allowRemove`. So its a cleaner API when you state what you are enabling/disabling in either case as an action, not as an internal component.

kyleledbetter:
ok maybe it should be `chipAddition` and we can also add a flag to allow the removal of chips `chipRemoval`

JoshSchoen:
Sure, I can make the changes. To be clear when you say chipRemoval does that mean when the user clicks on one of the chip list items when typing in autocomplete is the expected behavior to remove from the selected/filtered list rather than add to the selected/filtered list? If so does there need to be a toggle to add them back?

JoshSchoen:
@kyleledbetter @emoralesb05 I pushed the change to hide the `.mat-input-underline` and `allowAdd` to `chipAddition` I figured it's probably best to get that out of the way and do another PR for `chipRemoval` Let me know about the question in my last comment about expected `chipRemoval` behavior and I can take that on as well. Let me know if there is anything else. Thx!

emoralesb05:
@JoshSchoen ill create an issue later with the specifications of how `chipRemoval` should work. 

Merging this for now. Nice work dude! 💥

Issue #511, opened 2017-04-18
##Feature Request ## Hide data table column


arunanamal:
Hi,
Is there way to hide table column, but need to capture that value when row selected.

Thanks

emoralesb05:
Not yet, but we are gonna add this feature soon. Possibly by having a boolean property in the `ITdDataTableColumn`

arunanamal:
Thank you....

kyleledbetter:
found a good example http://iamisti.github.io/mdDataTable/

![image](https://cloud.githubusercontent.com/assets/867883/25683485/3b076c4a-3011-11e7-8d62-b15d5e3e7cbe.png)

arunanamal:
Hi Kyle Ledbetter,
Thanks for the info., but this is done by using some older version of Angular (AngularJS)?
How to apply this for Angular?

Thank you..

kyleledbetter:
We're gonna build the feature in Covalent. That was just inspiration for us :)

arunanamal:
Hi,
This is very basic question. How do I apply above changes for the existing
project?
Could you please tell me steps to follow to update above changes

Thanks

emoralesb05:
You can install it from the `nightly` repo to start testing while it gets released

`npm install https://github.com/Teradata/covalent-nightly.git`

arunanamal:
Thank you...

PR #548, opened 2017-05-03
feat: Data Table: Searchable and Unsearchable columns

### LINKED ISSUE #513: Data Table: Searchable and unsearchable columns.
Is it possible to flag a column to not to be searchable?
If yes, how do I do that.

jeremysmartt:
## Description

Added notsearchable property on columns to keep certain columns from being searched on in the filterData method. Also added unit tests for table and updated documentation.

Closes #513

### What's included?

- modified:   src/app/components/components/data-table/data-table.component.html
- modified:   src/app/components/components/data-table/data-table.component.ts
- modified:   src/platform/core/data-table/data-table.component.ts
- modified:   src/platform/core/data-table/services/data-table.service.ts
- new file:   src/platform/core/data-table/data-table.component.spec.ts

#### Test Steps

- [x] checkout branch
- [x] Go to http://localhost:4200/#/components/data-table
- [x] Go to the "Data Table with components"
- [x] Click the Search magnify glass
- [x] type in 392 see no results
- [x] delete 392 from search box
- [x] Click the TOGGLE SEARCHING CALORIES COLUMN button
- [x] See 1 result row in the table
- [x] Make sure documentation makes sense

- [x] `npm run lint` passes.

non-searchable-columns

### DIFF
--- a/src/app/components/components/data-table/data-table.component.html
+++ b/src/app/components/components/data-table/data-table.component.html
@@ -328,7 +328,13 @@
 
               filter(): void {
                 let newData: any[] = this.data;
-                newData = this._dataTableService.filterData(newData, this.searchTerm, true, this.columns);
+                let nonSearchAbleColumns: string[] = this.columns
+                .filter((column: ITdDataTableColumn) => {
+                  return (typeof column.filter !== undefined && column.filter === false);
+                }).map((column: ITdDataTableColumn) => {
+                  return column.name;
+                });
+                newData = this._dataTableService.filterData(newData, this.searchTerm, true, nonSearchAbleColumns);
                 this.filteredTotal = newData.length;
                 newData = this._dataTableService.sortData(newData, this.sortBy, this.sortOrder);
                 newData = this._dataTableService.pageData(newData, this.fromRow, this.currentPage * this.pageSize);
--- a/src/app/components/components/data-table/data-table.component.html
+++ b/src/app/components/components/data-table/data-table.component.html
@@ -289,7 +289,7 @@
               columns: ITdDataTableColumn[] = [
                 { name: 'sku', label: 'SKU #', tooltip: 'Stock Keeping Unit' },
                 { name: 'item', label: 'Item name' },
-                { name: 'price', label: 'Price (US$)', numeric: true, format: v => v.toFixed(2), notsearchable: true },
+                { name: 'price', label: 'Price (US$)', numeric: true, format: v => v.toFixed(2), filter: true },
               ];
 
               filteredData: any[] = this.data;
--- a/src/app/components/components/data-table/data-table.component.html
+++ b/src/app/components/components/data-table/data-table.component.html
@@ -342,22 +342,26 @@
   </md-card-content>
   <md-divider></md-divider>
   <md-card-actions>
-    <button md-button color="primary" (click)="toggleSelectable()" class="text-upper">
-      Toggle Selectable
-      ({{selectable ? 'ON' : 'OFF'}})
-    </button>
-    <button md-button color="primary" (click)="toggleMultiple()" [disabled]="!selectable" class="text-upper">
-      Toggle Multiple
-      ({{multiple ? 'ON' : 'OFF'}})
-    </button>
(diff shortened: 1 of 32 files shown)


kyleledbetter:
love the PR, could we name the input something more simple like `filter="false"` and true is the default?

emoralesb05:
I feel that this is more a design pattern than an enhancement.. not sure about this yet.

jeremysmartt:
Right at first I thought maybe a design pattern, but then actually I don't think it is.  I think the ability to pass in the columns and get hits based on properties will continue to be needed.  For example, probably want to do the same thing when a column is hidden:
https://github.com/Teradata/covalent/issues/511

emoralesb05:
The use case is legit, just debating if it should be part of the interface or applied in a different way since this seems more of an sorting implementation thing.

In both cases, there is stuff to be done in `TdDataTableService` for sure since thats a sorting implementation.

jeremysmartt:
@kyleledbetter I updated the name of the input to be filter="false" and true is the default.

kyleledbetter:
since we have a new batch of filtering/hiding/showing features coming, thinking we can do a new demo since that current demo is getting a little convoluted

kyleledbetter:
maybe not another demo, but maybe update the current complex datatable demo to use slide toggles for options instead of md-buttons, i'll take a stab

thomaslennan:
Another framework provides the ability to define both global and individual filters. Wondering if this PR should really be addressed by being able to filter on a per column basis. I, as a user, would find it confusing that I saw data in a column of the table that wasn't visible after input of some global filter value that matched the value that saw in that column. Alternatively, I'd want some visual indicator that the particular column isn't participating in the filter?

src/platform/core/data-table/services/data-table.service.ts
```diff
@@ -13,14 +13,16 @@ export class TdDataTableService {
    *
    * Searches [data] parameter for [searchTerm] matches and returns a new array with them.
    */
-  filterData(data: any[], searchTerm: string, ignoreCase: boolean = false): any[] {
+  filterData(data: any[], searchTerm: string, ignoreCase: boolean = false, nonSearchAbleColumns?: string[]): any[] {
```
emoralesb05: Should we just call this `excludedColumns`?

src/platform/core/data-table/data-table.component.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+import 'hammerjs';
+import { Component } from '@angular/core';
+import { By } from '@angular/platform-browser';
+import { TdDataTableComponent, ITdDataTableColumn } from './data-table.component';
+import { TdDataTableService } from './services/data-table.service';
+import { CovalentDataTableModule } from './data-table.module';
+import { NgModule, DebugElement } from '@angular/core';
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+
+describe('Component: TdPagingBarComponent', () => {
```
emoralesb05: should be `'Component: DataTable'`.

src/app/components/components/data-table/data-table.component.html
```diff
@@ -328,7 +328,7 @@ <h4 class="md-subhead">Paging Bar / Search Box / Sortable components</h4>
 
               filter(): void {
                 let newData: any[] = this.data;
-                newData = this._dataTableService.filterData(newData, this.searchTerm, true);
+                newData = this._dataTableService.filterData(newData, this.searchTerm, true, this.columns);
```
emoralesb05: Demo needs to be updated with latest `API` changes.

src/platform/core/data-table/services/data-table.service.ts
jeremysmartt: updated

src/platform/core/data-table/data-table.component.spec.ts
jeremysmartt: updated

src/app/components/components/data-table/data-table.component.html
jeremysmartt: updated

Issue #546, opened 2017-05-03
[feature request] File Upload: add custom form data


farkidev:
This repository's issues are reserved for feature requests and bug reports.

### Do you want to request a *feature* or report a *bug*?
*feature*

#### What is the expected behavior?

A key value pair or FormData parameter should be added to IUploadOptions where we can pass custom form data e.g. file type, comment, etc.

#### What is the motivation / use case for changing the behavior?

I now we can headers but I think these kind of data doesn't belong to the header

##### Other information
If you are interested I can do it and create a pull request.

emoralesb05:
Sure, really interested in a PR with that 😄

fpserp:
Today I made the necessary changes  (this is my work account) but when I run ng serve --aot I get a JavaScript heap out of memory. (Everything else is working including ng build --aot.) There is an open [issue](https://github.com/angular/angular-cli/issues/5618) about it at the cli repo.
Should I create a PR with the my new feature branch anyway?

emoralesb05:
Sure, PR away and we can start discussing this kinds of things in the PR. 😄

PR #580, opened 2017-05-10
feat(file-service): send additional form data with TdFileService 

fpeterdev:
## Description

This is a feature requested from #546. (I created the issue but another account)

With a new optional  `formData` property in `IUploadOptions` we can send additional form data when using `TdFileService`.
There were no tests for the `TdFileService` so I created two test cases: one with additional headers and data and one without.

### What's included?

- new `formDat` property to `IUploadOptions`
- add elements to `formData` in the service
- new tests for `TdFileService`
- update for README.md file

I got a JavaScript heap out of memory error just as described in this [issue](https://github.com/angular/angular-cli/issues/5618)
- [x] `npm run lint` passes.


emoralesb05:
Awesome! Thanks for the PR! Will have the team check it out.

src/platform/core/file/file-upload/README.md
```diff
@@ -83,14 +83,15 @@ export class MyModule {}
...
+Recieves as parameter an object that implements the [IUploadOptions] interface. *Note: since the file is sent by the name 'file' you can't use the string 'file' as a key in [additionalFormData]*
 
 ```typescript
 interface IUploadOptions { 
   url: string; 
   method: 'post' | 'put'; 
   file: File;
-  headers?: {[key: string]: string} 
+  headers?: {[key: string]: string};
+  additionalFormData?: {[key: string]: string};
```
emoralesb05: Would it be better to make it work as one or the other? 

Like If i use `file` then `additionalFormData` wont be taken into account, but if i dont use file then i can use `additionalFormData`? With this i think we can call it `formData`, have the `file` string be part of it too.

fpeterdev: You are right, it's more straightforward that way.
To make that work `file` has to be optional to. What should happen when the user don't use neither `file` nor `code`? Throw an Error with the message "For [IUploadOptions] you have to set either the [file] or the [formData] property."?

emoralesb05: Yeah, we can though an error if both arent provided. 👍

emoralesb05:
Awesome! Looks good, gonna merge this to include it in beta.4 😄 

Thanks dude!

PR #594, opened 2017-05-15
chore(docs): make covalent homepage look right in IE11

mcstepp:
## Description
Even though IE11 isn't officially supported by Covalent, angular-material supports IE11 and all the components work in IE11. However, the Covalent home page doesn't really look that great in IE11. We want to show off Covalent at an upcoming demo, and stakeholders want to see that Covalent is compatible with IE11 so that we can continue to use it in TDAA applications. 

This change makes the Covalent homepage work and display well in IE11.

Before:
covalent_homepage_ie11_before

### What's included?

- .../app/components/home/home.component.html
- .../app/styles.scss (not needed, just picked up a white space change)

#### Test Steps

View in IE 11

- [x] `npm run lint` passes.

After:
covalent_homepage_ie11_after

### DIFF
--- a/src/app/components/home/home.component.html
+++ b/src/app/components/home/home.component.html
@@ -57,9 +57,10 @@
         <md-divider *ngIf="!last" md-inset></md-divider>
       </ng-template>
     </md-nav-list>
-    <div layout="row" layout-wrap hide-xs hide-sm class="push-sm">
+    <div>
+    <div layout="row" hide-xs hide-sm class="push-sm">
       <ng-template let-item let-last="last" ngFor [ngForOf]="items">
-        <div flex-md="50" flex-gt-md="25" layout="column">
+        <div flex-md="50" flex-gt-md="25" layout="row">
           <md-card href [routerLink]="[item.route]" class="md-card-colored" flex>
             <md-toolbar class="bgc-{{item.color}} " layout="column" layout-align="center center" layout-padding>
               <md-icon class="md-icon-font-xl tc-white">{{item.icon}}</md-icon>
@@ -74,7 +75,7 @@
         </div>
       </ng-template>
     </div>
-    <md-divider></md-divider>
+    </div>
     <md-card-actions>
       <div layout="row">
         <a href="https://gitter.im/Teradata/covalent" target="_blank" md-button color="primary" class="text-upper">Gitter Chat</a>
--- a/src/app/components/home/home.component.html
+++ b/src/app/components/home/home.component.html
@@ -57,8 +57,7 @@
         <md-divider *ngIf="!last" md-inset></md-divider>
       </ng-template>
     </md-nav-list>
-    <div>
-    <div layout="row" hide-xs hide-sm class="push-sm">
+    <div layout="row" layout-wrap hide-xs hide-sm class="push-sm">
       <ng-template let-item let-last="last" ngFor [ngForOf]="items">
         <div flex-md="50" flex-gt-md="25" layout="row">
           <md-card href [routerLink]="[item.route]" class="md-card-colored" flex>
@@ -75,7 +74,7 @@
         </div>
       </ng-template>
     </div>
-    </div>
+    <md-divider></md-divider>
     <md-card-actions>
       <div layout="row">
         <a href="https://gitter.im/Teradata/covalent" target="_blank" md-button color="primary" class="text-upper">Gitter Chat</a>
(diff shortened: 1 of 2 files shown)


src/app/components/home/home.component.html
```diff
@@ -258,4 +258,4 @@ <h3 md-line>Feature Requests & Bugs</h3>
       <md-icon class="md-icon-ux" svgIcon="assets:teradata-ux"></md-icon>
     </div>
   </td-layout-footer>
-</td-layout-nav>
+</td-layout-nav>
```
aciccarello: Did this remove the EOL character?

mcstepp: I'm not sure. I grabbed the raw code to make sure I didn't leave out any of the attributes after a premature commit and re-applied my fix to the raw code.

kyleledbetter:
It's super weird that this fixes IE11. The cards need to be `layout="column"` bc they're stacking vertically and `layout="row"` would make the card toolbar and body be side by side. Does this not break the other browsers?

emoralesb05:
Yeah, im wondering the same thing.. maybe the real issue is missing rules from the `flex` classes for IE11 support..

Blankinfinity:
To get the main page to look correct i just changed the `push-sm` class to have a `height` of `300px`, this seems to have worked for me and i haven't noticed any knock on effects from that.

kyleledbetter:
Right but push-sm is a global utility for small margins and that would blow up everyone's app :)
The only way to get this solved properly is by getting all flex layout classes working fully in IE11. We'll do some tests for the next release but no guarantees

mcstepp:
>It's super weird that this fixes IE11. The cards need to be layout="column" bc they're stacking vertically and layout="row" would make the card toolbar and body be side by side. Does this not break the other browsers?

The cards still "column" layout. This is the wrapper element around the card itself: `<div flex-md="50" flex-gt-md="25" layout="row">`
                       
One of the IE 11 issues is that flex item children don't respect the min-height of the flexbox container parent; it can't properly calculate how much space the flex children have to flex about. One of the fixes for that is to wrap the flexbox container with another flexbox container. I don't know why nested flexbox doesn't have this issue, but it doesn't. You've already got the flex box wrapper, so that's fine. 

The reason why the direction should change though is because flexbox direction `column` has some weird behavior where it collapses the height to 0 in IE11. This is what we're seeing in the "before" screenshots - the cards aren't fitting inside the wrapper. See here: https://connect.microsoft.com/IE/feedback/details/802625/min-height-and-flexbox-flex-direction-column-dont-work-together-in-ie-10-11-preview

And since the `md-card` is the only child of that wrapper element, for other browsers it doesn't matter what direction it is in. Row vs column on the wrapper element produces the same results because there's one one flex item child - the card itself. 

```
<div flex-sm="50" flex-md="50" flex-gt-md="25" layout="row">
    <md-card href [routerLink]="[item.route]" class="md-card-colored" flex>
        <md-toolbar class="bgc-{{item.color}} " layout="column" layout-align="center center" layout-padding>
            <md-icon class="md-icon-font-xl tc-white">{{item.icon}}</md-icon>
        </md-toolbar>
        <md-card-content>
            <div layout="column" layout-padding>
                <div class="md-subhead">{{item.title}}</div>
            </div>
        </md-card-content>
    </md-card>
</div>
```

Blankinfinity:
Ok, i retract my previous comment ;) Would setting a `min-height `on the nested div beneath the one with the `push-sm` class break anything, as this also works. Excuse me if i'm being stupid or ignorant but it seems like a lot less hassle.

image

mcstepp:
A lot less hassle is merely changing the layout direction on the wrapper element which is what this PR does. It doesn't require extra non-semantic elements either.

Works in Edge: 
screen shot 2017-05-16 at 10 52 41 am
screen shot 2017-05-16 at 10 53 12 am

Works in Chrome:
screen shot 2017-05-16 at 10 54 57 am
screen shot 2017-05-16 at 10 55 24 am

Works in Safari:
screen shot 2017-05-16 at 10 55 55 am
screen shot 2017-05-16 at 10 56 19 am

I didn't test it in Firefox (I don't have firefox on my laptop because TDAA doesn't support Firefox)

emoralesb05:
Yeah, thats seems to be correct @mcstepp . Since it only has one child it doesnt matter if its flex column or row in this case.

PR #606, opened 2017-05-19
feat(dialog): added component

jotatoledo:
added ConfirmAgainstValueDialog component 
feat(dialog): added method in service

* added openConfirmAgainstValue: creates a new type of dialog instance
* added new interface to define the config for the new functionality
feat(dialog): registered component in module

registered TdConfirmAgainstValueDialogComponent


emoralesb05:
Thanks for the PR man! Will start checking it out tomorrow 😄

jotatoledo:
no worries, I havent had the time to check what caused the problem on the travis run

emoralesb05:
Been wondering if it might be simpler to just add an extra property into the `prompt` dialog.. since in essence its the same thing with an extra validation... since it seems like lots of duplicated code between this and `prompt`.

if the property is undefined it could act as it normally does, but if it has a value, then it could validate against that. Probably a `validateAgainst` property

jotatoledo:
noticed that to, but I wasnt sure if u actually wanted to extend the behaviour of the prompt dialog or not :smile: 
I will refactor the changes

emoralesb05:
haha cool. thanks man!

PR #619, opened 2017-05-24
fix(data-table): block (rowClick) event when clicking on checkbox (closes #611)

### LINKED ISSUE #611: [ BUG ] datatable rowselect trigger rowclick even too
#### Bug Report
BUG (Not correct behavior)

i've tried rowclick event in datatable but got some bug
the rowselect method is not working properly,
so when you click the checkbox it will trigger the rowclick method instead of rowselect method
you can see the example in the documentation too

emoralesb05:
even though clicking on the checkbox is still clicking on the row, we need to block this since people want to navigate when clicking on row and select when clicking on checkbox.

https://github.com/Teradata/covalent/issues/611

### What's included?

- block `(rowClick)` event when click event target is `md-pseudo-checkbox`
- unit test for this particular use case

#### Test Steps

- [ ] `ng serve`
- [ ] Go to data-table demo http://localhost:4200/#/components/data-table
- [ ] Go to last demo and `enable` clickable
- [ ] See that now it only shows the dialog when clicking on the row and not when toggling the checkboxes.

- [ ] `npm run lint` passes.

### DIFF
--- a/src/platform/core/data-table/data-table.component.html
+++ b/src/platform/core/data-table/data-table.component.html
@@ -30,7 +30,7 @@
         [tabIndex]="isSelectable ? 0 : -1"
         [class.mat-selected]="(isClickable || isSelectable) && isRowSelected(row)"
         *ngFor="let row of data; let rowIndex = index"
-        (click)="handleRowClick(row, $event, rowIndex)"
+        (click)="handleRowClick(row, $event)"
         (keyup)="isSelectable && _rowKeyup($event, row, rowIndex)"
         (keydown.space)="blockEvent($event)"
         (keydown.shift.space)="blockEvent($event)"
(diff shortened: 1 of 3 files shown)


jeremysmartt:
Made an update to this.  Instead of looking for a tag name (MD-PSEUDO-CHECKBOX) to know to stop the event propogation, I added an attribute to the tag to look for. Looking for the tag name could be problematic in that what if someone wants to put their own md-pseudo-checkbox on a cell and doesn't want the event to be stopped.  And in fact anyone could now use this new attribute stopEventPropogation="true" on any of their own cells if they wanted to stop the event from happening for any number of reasons.

emoralesb05:
Awesome, its more flexible this way and not dependent of the element.

PR #639, opened 2017-05-30
feat(chips): new `[stacked]` input to make chips stack vertically.

emoralesb05:
New input to be able to stack chips vertically or horizontally depending on its boolean value.

`[stacked]` will be false by default.

### What's included?

- `[stacked]` input with demo, docs and usage.
- unit tests for `[stacked]`

#### Test Steps

- [x] `ng serve`
- [x] Go to http://localhost:4200/#/components/chips
- [x] put `stacked` in any chips demo.

- [ ] `npm run lint` passes.

screen shot 2017-05-29 at 7 14 47 pm


jeremysmartt:
should add filtering to demo

jeremysmartt:
should add ESC to make autocomplete list go away

richavyas:
Found a bug. When a chip is selected say `stepper` from the stacked list:
- it should go away from the list so that user cannot select it again.
- if the same option is again clicked, `show stacked list on mouse-click` stops working and the stacked list disappears.

emoralesb05:
Not a bug haha 😄 , but we already discussed it and we need to make the list filter in the demo too.

emoralesb05:
Actually nvm, the filtering is not a bug, but the clicking on a existing is a bug. Already did a fix for that while adding the `ESC` a11y for the input

PR #643, opened 2017-05-30
chore(common): add common module setup in directives and pipe docs. (closes #634)

### LINKED ISSUE #634: Documentation: missing import explanation for custom pipes
There is no description of how to import the custom pipes in https://teradata.github.io/covalent/#/components/pipes

emoralesb05:
## Description

https://github.com/Teradata/covalent/issues/634

There was no documentation on how to include our pipes and directives, so we needed to add that into both views.

Also removing `max`/`min` validators since angular already has them by default with the same selectors.

### What's included?

- chore(): remove max/min validators since angular already has them
there will be no breaking change since the selectors are the same

- chore(common): add common module setup in directives and pipe docs

#### Test Steps

- [x] `ng serve`
- [x] Go to http://localhost:4200/#/components/pipes
- [x] See setup
- [x] Go to http://localhost:4200/#/components/directives
- [x] See setup and no mention on max/min

- [ ] `npm run lint` passes.


kyleledbetter:
should we link to the Angular docs on min/max and other Pipes, etc they have like Date?

kyleledbetter:
the only angular docs i found reference `minlength="4" maxlength="24"` VS what we had `min="4" max="24"` would that be a breaking change for us?

got that info from here (should we add links to this?):
https://angular.io/docs/ts/latest/api/forms/index/MinLengthValidator-directive.html
https://angular.io/docs/ts/latest/api/forms/index/MaxLengthValidator-directive.html

PR #669, opened 2017-06-07
fix(steps): change internal class to mat-inactive

emoralesb05:
## Description

Makes more sense to call it `mat-inactive` since it just colors it and helps out in unit testing when trying to figure out if a step is disabled or inactive.

- [ ] `npm run lint` passes.


jerryorta-dev:
approved

PR #706, opened 2017-06-22
Update package.json

agborkowski:
update material to beta.7


emoralesb05:
Thanks for the PR, but there are a couple of issues in the mixins when upgrading to beta.7. 

We also have to test all components to see if they work correctly when upgrading to a new material version since we are deeply couple to them.

PR #708, opened 2017-06-23
feat(dynamic-forms): add custom validation capability and error message template

jerryorta-dev:
# Proof of Concept - Custom validation for dynamic-forms

## Description

Three custom validation scenarios are implemented to accommodate project needs and covalent  scalability. 

- Simple config object
- Pure function in the form `ValidatorFn` from `@angular/forms`
- Validators migrated to the covalent platform

### Extending the `dynamic-forms` API

```
textElements: ITdDynamicElementConfig[] = [{
    name: 'input',
    type: TdDynamicElement.Input,
    required: false,
    customValidators: [
          // Add your custom validators here
     ]
  }]
```

A property named `customValidators` is added to `ITdDynamicElementConfig` which contains an array of custom validation configs ( `ITdCustomValidate` ) or pure functions with the signature and architecture of `ValidatorFn` from `@angular/forms`.

`ITdCustomValidate`  configs and `ValidatorFn` functions contained in the `customValidators` array may be mixed and in any order.

#### Simple Config Object

To ease the development of simple and quick custom validation, use the simple `ITdDynamicElementConfig`  config. The benefit is quick and easy implementation. The limitation is the message output needs to be static ( no input values added to message ).

```
    export interface ITdCustomValidate {
        name: string;
        message: string;
        predicateFn(value: any): boolean ; // return true if valid
    }
```
 
And implemented as: 

```
textElements: ITdDynamicElementConfig[] = [{
    name: 'input',
    type: TdDynamicElement.Input,
    required: false,
    customValidators: [
          // Add your custom validators here
         
           <ITdCustomValidate>{
                name: 'containsNumber',
                message: 'Need a number',
                predicateFn: (value: any): boolean => {
                   return /\d/i.test(value); // Return true if valid
                 },
             },

     ]
  }]
   ```

#### ValidatorFn 

A pure `ValidatorFn` type function provides the lowest level development of a validator function. The developer is required to adhere to the `ValidatorFn` architecture from `@angular/forms`, and to return an error object ( if inValid ) of the interface `ITdCustomError`.

```
textElements: ITdDynamicElementConfig[] = [{
    name: 'input',
    type: TdDynamicElement.Input,
    required: false,
    customValidators: [
          // Add your custom validators here

         // Contain at least one special character
         <ValidatorFn>(control: AbstractControl): { [key: string]: any } => {

             let validCharacters: string[] = ['!', '@', '#', '$', '%'];
             let regex: RegExp = new RegExp('[' + validCharacters.join('').toString() + ']', 'g');
        
             let isValid: boolean = regex.test(control.value);

             let error: ITdCustomError = {
                'containSpecialCharacters': {
                     message: 'Need at least one special character (' + validCharacters.join(', ').toString() + ')',
                },
            };

        return !isValid ? error : undefined;
      },

     ]
}]
```

#### Migration to covalent

Architecture is provided to easily add validations to covalent, extending the dynamic-forms API.

```
textElements: ITdDynamicElementConfig[] = [{
    name: 'input',
    type: TdDynamicElement.Input,
    required: false,

   myAwesomeValidatorInCovalent: true, // <-- migrated from project

  }];
```

#### Test Steps

- [ ] `git clone feature/jo-poc-custom-validation-565`
- [ ] `ng serve`
- [ ] navigate to [http://localhost:4200/#/components/dynamic-forms](http://localhost:4200/#/components/dynamic-forms)

- [ ] `npm run lint` passes.

### CODE REVIEW

File: src/platform/dynamic-forms/dynamic-elements/dynamic-input/dynamic-input.component.html [hunk 1]
```diff
@@ -10,9 +10,8 @@
             [attr.max]="max"
             flex>
     <md-hint align="start">
-      <span class="tc-red-600">
-        <span *ngIf="control.hasError('max')">Max: {{max}}</span>
-        <span *ngIf="control.hasError('min')">Min: {{min}}</span>
+      <span class="tc-red-600"  *ngIf="control.errors && control.dirty">
+        <span *ngFor="let error of errorMessages; last as isLast">{{error}} <span *ngIf="!isLast">, </span></span>
```
emoralesb05: I think we can either show all the errors like you have and put a `.` as separator since the material example does something like this https://material.io/guidelines/patterns/errors.html

Its either that, or give it like order priority to only show one message at a time ( which i'm still debating)

src/platform/dynamic-forms/dynamic-elements/dynamic-input/dynamic-input.component.ts
```diff
@@ -29,4 +29,42 @@ export class TdDynamicInputComponent extends AbstractControlValueAccessor implem
...
+      if (this.control.errors) {
+        for (let key in this.control.errors) {
+
+          let errorMessage: string = undefined;
+
+          // Angular Standard
+
+          // Not checking required
+
+          if (key === 'min') {
```
emoralesb05: Might also be cool if we can also pass `min`/`max`/`required` validators like this to provide custom error message and keep the existing properties only to modify the native input by either showing the `*` with required and blocking the up/down number addition with min/max.

src/platform/dynamic-forms/validators/dynamic-forms.validators.tdContainsUppercaseCharacter.ts
```diff
@@ -0,0 +1,29 @@
+import { AbstractControl, ValidatorFn, FormGroup } from '@angular/forms';
+import { ITdCustomError } from './dynamic-forms.validators.interfaces';
+
+export function tdContainsUppercaseCharacter(control: AbstractControl): { [key: string]: any } {
```
emoralesb05: Lets remove this one, and if we want to show an example we can just add it as a `Validator`, so its not part of the interface.

src/app/components/components/dynamic-forms/dynamic-forms.component.ts
```diff
@@ -104,4 +114,117 @@ export class DynamicFormsDemoComponent {
...
+    // Covalent validator migrated from other project
+    tdContainsUppercaseCharacter: true,
+    
+    customValidators: [
+
+      // Contain at lease one number
+      <ITdCustomValidate>{
+        name: 'containsNumber',
+        message: 'Need a number',
+        predicateFn: (value: any): boolean => {
```
emoralesb05: Would i be able to pass a Validator from angular or covalent? 

e.g.

`Validator.required`?

src/platform/dynamic-forms/dynamic-elements/dynamic-input/dynamic-input.component.html
jerryorta-dev: I think priority will be very hard to determine since validation is custom. I do like the idea of only displaying one error at a time, so what do you think of just showing one error at a time?

src/platform/dynamic-forms/dynamic-elements/dynamic-input/dynamic-input.component.ts
jerryorta-dev: I can work on that. What is "blocking the up/down number addition with min/max." Is that the up/down button the browser adds with number fields?

src/platform/dynamic-forms/validators/dynamic-forms.validators.tdContainsUppercaseCharacter.ts
jerryorta-dev: Great. I was waffling on this and couldn't make a decision. 

This was just a PoC example, but do you think we would ever downstream custom validations from products if requested and approved?

src/app/components/components/dynamic-forms/dynamic-forms.component.ts
jerryorta-dev: That is exactly the intent as well.

src/platform/dynamic-forms/dynamic-elements/dynamic-input/dynamic-input.component.ts
emoralesb05: Yeah, basically the way it works now would basically add those attributes. But the real validation would be added via your new property with `Validators.required`, `Validators.max` and `Validators.min` and a message

src/platform/dynamic-forms/validators/dynamic-forms.validators.tdContainsUppercaseCharacter.ts
emoralesb05: Yeah, i would want to keep the dynamic forms agnostic of which validations exist and leverage this way of adding them, so from a devs perspective they would just add a new validation entry in the array or create a composite validation with `Validators.compose` and pass it.

src/platform/dynamic-forms/dynamic-elements/dynamic-input/dynamic-input.component.html
emoralesb05: Yeah, its a hard debate. 

We can either just keep it to the devs decision and just show all of them finishing the sentences with `.` or always show the first error from the array, which the array order would somewhat be the priority order.

e.g. if `n` errors are visible, we only show the first one.

whatever makes more sense.

src/app/components/components/dynamic-forms/dynamic-forms.component.ts
emoralesb05: :+1:

src/platform/dynamic-forms/dynamic-elements/dynamic-input/dynamic-input.component.html
jerryorta-dev: @emoralesb05 I want to refactor input error to use `<md-error>` to get closer to standards, as well show only one error at a time. Please let me know if you think this plan should be different.

src/platform/dynamic-forms/validators/dynamic-forms.validators.tdContainsUppercaseCharacter.ts
jerryorta-dev: removed

src/platform/dynamic-forms/dynamic-elements/dynamic-input/dynamic-input.component.ts
```diff
@@ -1,21 +1,14 @@
...
 import { AbstractControlValueAccessor } from '../abstract-control-value-accesor';
 
-export const INPUT_INPUT_CONTROL_VALUE_ACCESSOR: any = {
-  provide: NG_VALUE_ACCESSOR,
-  useExisting: forwardRef(() => TdDynamicInputComponent),
-  multi: true,
-};
-
 @Component({
-  providers: [ INPUT_INPUT_CONTROL_VALUE_ACCESSOR ],
```
jerryorta-dev: This was not triggering <md-error> to show, though I could see the error object in the Component code.

src/platform/dynamic-forms/dynamic-elements/dynamic-input/dynamic-input.component.html
```diff
@@ -3,16 +3,13 @@
     <input #elementInput
             mdInput
             [(ngModel)]="value"
+            [formControl]="control"
```
jerryorta-dev: Triggers <md-error> to show errors.

src/app/components/components/dynamic-forms/dynamic-forms.component.ts
jerryorta-dev: I have to check the error for each Validator, we were already doing this for min and max, but need to extend for all Angular Validators.

jerryorta-dev: Now able to add Angular Validators

emoralesb05:
I updated the PR with a simpler interface and an error template to provide error messages, this way we are future proofing for i18n.

For now im using a normal div to show/hide the errors, but when material releases their `form-field` element we might be able to leverage and return to use `md-error` again.

PR #725, opened 2017-07-01
Change viewContainerRef from optional to required

eddiejaoude:
## Description

> Note: if no [ViewContainerRef] is provided, [TdDialogService] will throw an error.

On this line further up the [file](https://github.com/Teradata/covalent/blob/develop/src/app/components/components/dialogs/dialogs.component.html#L18) it states it will throw an error if not provided, then further down in the example it says it is `OPTIONAL`. It seems it should be `REQUIRED`.

### What's included?

- Code example comment change

#### Test Steps

N/A

N/A

N/A

### DIFF
--- a/src/app/components/components/dialogs/dialogs.component.html
+++ b/src/app/components/components/dialogs/dialogs.component.html
@@ -45,7 +45,7 @@
             this._dialogService.openAlert({
               message: 'This is how simple it is to create an alert with this wrapper service.',
               disableClose: true | false, // defaults to false
-              viewContainerRef: this._viewContainerRef, //OPTIONAL
+              viewContainerRef: this._viewContainerRef, // REQUIRED
               title: 'Alert', //OPTIONAL, hides if not provided
               closeButton: 'Close', //OPTIONAL, defaults to 'CLOSE'
             });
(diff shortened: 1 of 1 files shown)


emoralesb05:
Its actually the other way around.. before it was required, and now its not..

We just forgot to remove the `Note` from the docs.. can you remove that note and return it to `optional`?

eddiejaoude:
> We just forgot to remove the Note from the docs.. can you remove that note and return it to optional?

Sure, I can remove it. But, just want to double check that is correct, as when I didn't provide it I got an error?

emoralesb05:
Yeah, you dont need to provide it anymore, only if you want to. 

Unless you are using an old version of covalent

eddiejaoude:
Ok. 

I thought I was using the latest one ...

```json
"@covalent/core": {
      "version": "1.0.0-beta.5-1",
      "resolved": "https://registry.npmjs.org/@covalent/core/-/core-1.0.0-beta.5-1.tgz",
      "integrity": "sha1-UD2Ba8s2vxZKOIhwrXTM0kCxKEY="
    }
```

emoralesb05:
Yeah you are, is that giving you an error?

eddiejaoude:
Yep I was. I am no longer using the dialog. I will test again shortly.

PR #728, opened 2017-07-03
make progress based on upload API

rafal83:
Make progress upload based on upload, not need backend compatible


emoralesb05:
Thanks, now its pointing to develop 😄

PR #752, opened 2017-07-09
fix(change-log): td-layout-manage-list example typo

davidgabrichidze:


emoralesb05:
Thanks, dunno how i missed that haha

PR #784, opened 2017-07-24
fix(chips): close manually autocomplete menu on mousedown

### LINKED ISSUE #783: When there are multiple chips on the same page... Only one dropdown list should be visible at a time
#### Bug Report

If you have several chips inputs on the same page... then clicking above one of them will trigger the top one to show its menu while not closing the bottom one. 

I was able to reproduce this issue on the demo page.
1. Click on the bottom chips component and get  the menu to apear.
2. Click on the top chips component where its highlighted in the screenshot and enter something that would cause the popup to appear

image

image

Only one dropdown menu should be visible at a time.

emoralesb05:
since we are stopping the propagation of the click event for chips, this causes some issues with autocomplete since its listening to it

#### Test Steps

- [ ] `ng serve`
- [ ] Try replicating it with the info in #783 

- [ ] `npm run lint` passes.

closes #783

### DIFF
--- a/src/platform/core/chips/chips.component.ts
+++ b/src/platform/core/chips/chips.component.ts
@@ -714,7 +714,7 @@ export class TdChipsComponent extends _TdChipsMixinBase implements ControlValueA
    */
   private _watchOutsideClick(): void {
     if (this._document) {
-      this._outsideClickSubs = RxChain.from(fromEvent(this._document, 'click')).call(filter, (event: MouseEvent) => {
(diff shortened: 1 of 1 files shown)


ankurbrdwj:
was this merged to beta 8 or above its not there ?

emoralesb05:
This has been there for quite a while, if you still have an issue like this, can you please open a new issue with steps to replicate it?

PR #786, opened 2017-07-25
Data Table nested sorting

### LINKED ISSUE #704: Bug Report: Sorting with nested names in TdDataTableService.sortData
Hi,
Im using covalent 1.0.0-beta.5 and tried out the td-data-table with sorting capabilities. 
The problem I have is when we present data from a nested hierarchy (see example below) the sorting does not work.

Example:
`
columns: ITdDataTableColumn[] = [
    {name: 'commitRef.commitTime', label: 'Updated', tooltip: 'Updated', sortable: true, format: (value) => this.datePipe.transform(value, 'yyyy-MM-dd HH:mm:ss')},
    {name: 'commitRef.committer.name', label: 'Changed by', tooltip: 'Changed by', sortable: true}
  ];`

Then when calling sortData in TdDataTableService it doesn't retrieve the value from nested properties.

https://github.com/Teradata/covalent/blob/4544c1a5bcabb1fa3e60b6fe6c2f6e1a5215ab5c/src/platform/core/data-table/services/data-table.service.ts#L45

In my case above this would resolve into a["commitRef.commitTime"], where it actually should have been something like a["commitRef"]["commitTime"]

JoshSchoen:
## Enables nested sorting with the data table.
closes #704

Enabled the ability to sort by nested data values with the data table. Added an interface property called `sortBy` checks if it is defined otherwise uses the name property. 

Modified the `sortData` method within data-table.service.ts to allow nested sorting.
Also included a new unit test for the `srotData` method for data-table.service.ts

### What's included?

src/app/components/components/data-table/data-table.component.html
src/app/components/components/data-table/data-table.component.ts
src/platform/core/data-table/data-table-column/data-table-column.component.ts
src/platform/core/data-table/data-table.component.ts
src/platform/core/data-table/services/data-table.service.spec.ts
src/platform/core/data-table/services/data-table.service.ts

#### Test Steps

- [ ] Checkout Branch
- [ ]  ng serve
- [ ] Scroll Down to Data Table sorting using nested objects
- [ ] Click on headers to sort the data for each column, initial state should be: "Two Levels, sorts on second level value" and sorted Ascending.

- [ ] `npm run lint` passes.

nested-data-table-sort


JoshSchoen:
@emoralesb05 let me know how you want to proceed with this.

src/app/components/components/data-table/data-table.component.html
```diff
@@ -394,7 +394,264 @@ <h4 class="md-subhead">Paging Bar / Search Box / Sortable components</h4>
...
+        <div layout="row" layout-align="start center" class="pad">
+          <span class="md-title">Title here</span>
+        </div>
+        <md-divider></md-divider>
+        <td-data-table
+          [data]="nestedFilteredData"
+          [columns]="nestedColumnsExample"
+          [sortable]="true"
+          [sortOrder]="sortOrder"
+          [sortBy]="nestedSortBy"
```
jeremysmartt: Why is this defined on the table?  The sortBy input that is?

Doesn't it just need to be on the column definition?

```
nestedColumnsExample: ITdDataTableColumn[] = [
   { name: 'predicted.nested.nested.nested', label: 'Four Levels', sortable: true, sortBy: 'predicted.nested.nested.nested.value' }, 
```

jeremysmartt: Or the other way around, defined on table but not column.  One or the other I guess?

JoshSchoen:
@jeremysmartt sorry for the delay been hectic this week. I will take a look and update.

JoshSchoen:
@jeremysmartt just updated the the html template and you were correct the sortBy was not needed. Let me know if you see anything else.

JoshSchoen:
@jeremysmartt I don't think we need this any longer. Not sure if this was available then but using let-row="row" with `tdDataTableTemplate` gives me access to what I think I needed back then and suitable for my need. I'll close this PR. Thanks!

PR #794, opened 2017-07-28
chore(): first iteration on kitchen sink for server-side rendering test script

stevenov7:
## Description

Created a package.json script to see if particular Covalent components will render successfully server side. If it fails the terminal error will indicate what issues need to be resolved. Currently the script tests a single Covalent component, `<td-message>`, just to get us started. The messaging component correctly fails. It fails because we are referencing `window` inside `@covalent/core` which is not available server side. Adding additional components to test in the future is straight forward. 

Note: I tested the script against the material button component and it successfully passed/rendered server-side.

### What's included?

- New script `npm run combat-training`

#### Test Steps

- [ ] `npm run reinstall`
- [ ] `npm run combat-training`
- [ ] Verify this error appears: 
```
covalent/node_modules/@covalent/core/core.umd.js:7
...
ReferenceError: window is not defined
...
```

- [ ] `npm run lint` passes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 # See http://help.github.com/ignore-files/ for more about ignoring files.
 
 # compiled output
-/dist
+/**/dist
 /deploy
 /tmp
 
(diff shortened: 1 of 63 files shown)


ElderJames:
Great work！

PR #812, opened 2017-08-12
feature(dialog): added validation logic to prompt/confirm dialogs

### LINKED ISSUE #602: Feature request: Confirmation dialog with extra functionalities
#### Feature Request

A confirmation dialog with the option to check the value of an input field in it against an injected value.
If the values match, then the `confirm `option is active. In other case, the `confirm `button is disabled.

The confirmation dialog will most likely be used for confirmation of delete actions.  An user that wants to `cancel `this action could still click the `confirm `option by mistake, leading to chaos and panic. As a security measure, implementing something like this would be really helpful. 

##### Other information

My basic idea is to create something similar to the confirmation dialog on github for a delete repository action.

### LINKED ISSUE #660: TdDialogService should take an optional validator as an input
#### Feature Request
TdDialogService should allow putting a validator while we enter the text.
I am using it to rename files and would need to disable the okay button in the case the name already exists.

jotatoledo:
## Description

Added an extra validation step in the prompt and confirm dialogs.

**Confirm dialog** 

The addition of this allows developers to create an extra validation step before being able to confirm some operation.

_Typical use case_: delete something

If the "something" to be deleted could cause mayor side effects (example: delete a repo in github), to confirm this through a simple click isnt safe enough.  In the case of github, a confirm dialog will be shown, in which the user needs to type in the name of the repo to delete. This can now be archived through this pr.

**Prompt dialog**

Developers can now filter what users type in, before allowing them to click confirm.

_Typical use case_: allow any value, except for the ones in an array.

To get a rough idea of what I tried to archieve, please take a look in the following plnkr:

http://plnkr.co/edit/mxVwIc7qRHinLjY6FEQa

Closes #602 
Closes #660

### What's included?

- Added property to `IConfirmConfig` interface
- Changed `openConfirm` and `openPrompt` methods in dialog service
- Changes to `ConfirmDialog` and `PromptDialog` components

#### Test Steps

Dont apply

- [ ] `npm run lint` passes.

### DIFF
--- a/src/platform/core/dialogs/confirm-dialog/confirm-dialog.component.html
+++ b/src/platform/core/dialogs/confirm-dialog/confirm-dialog.component.html
@@ -4,6 +4,12 @@
   </td-dialog-title>
   <td-dialog-content class="md-subhead tc-grey-700">
     {{message}}
+    <form *ngIf="validatorIsSet" #form="ngForm" layout="row" novalidate flex>
+      <md-input-container flex>
+        <input mdInput (keydown.enter)="$event.preventDefault(); form.valid && valid && accept()"
+          [(ngModel)]="input" name="input" required/>
+      </md-input-container>
+    </form>
   </td-dialog-content>
   <td-dialog-actions>
     <button md-button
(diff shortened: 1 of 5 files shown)


emoralesb05:
Awesome dude! Just wondering if the `confirm` dialog has to be changed since the `prompt` one basically achieves the same thing with the validation 🤔

jotatoledo:
In a certain way yes, they archive more or less the same but the APIs are different. The prompt dialog has a different return, the input itself, and in my opinion a different goal. The confirm only returns a yes/no, so the input isnt really relevant for the end result, its only an in-between step so to say.

Issue #891, opened 2017-09-29
Why is the min column width arbitrarily hard coded to 100px?


praveenpuglia:
I am trying to build a table where I want a column to be fixed at 60px. So I looked at the API and saw that I can pass the width as a number. Didn't work. 

Then I saw I could pass an object `{min: NUMBER, max: NUMBER}` for the column configuration and I hoped that would do it for me. No! It was always at least 100px.

I went here - https://github.com/Teradata/covalent/blob/develop/src/platform/core/data-table/data-table.component.ts#L972-L975 and I realized that any column declaring width less than 100 px is being set to 100px. Which is weird. 

Right now, I can't possibly set the width of any column narrower than 100px :(

kyleledbetter:
100 is the default, try setting width to something smaller than 100. To be clear, set width specially not min or max

praveenpuglia:
I tried that but even then I get a minimum of 100px.

kyleledbetter:
Can you post a stackblitz or plunker?

dubemarcantoine:
I noticed the same issue when I put in smaller values than 100px.
When investigating, I found another issue, but I don't know if it is related, so I will comment it here:
The `width: {min: 300, max: 500}` object when defining my columns, the max-width is applied as the min-width on the element.

Here is the element style from Google Chrome.
<img width="398" alt="screen shot 2017-09-29 at 12 45 57 pm" src="https://user-images.githubusercontent.com/11211814/31026440-7186afec-a514-11e7-8298-7a8ad987bc21.png">

I will try and continue on how to fix this if. It seems that the issue first came from this PR. [https://github.com/Teradata/covalent/pull/859](url). Before this, columns would expand by width and height naturally depending on the content in the grid.

praveenpuglia:
@dubemarcantoine is there a plunker template?

dubemarcantoine:
I did not do one yet, I'll do it later today/tomorrow after school!

praveenpuglia:
I am sorry. I wanted to tag @kyleledbetter :) Is there a plunker template i can use?

emoralesb05:
The `min`/`max` is indented to work that way, because table cells are really stubborn when trying to set the widths.. we calculate the needed width and set it to both min/max to ensure it works properly.

The 100px one was so we could have a min width, but i do agree 100px seems really random and you should be able to override it. A fix for this will be addressed https://github.com/Teradata/covalent/pull/895

PR #895, opened 2017-10-01
fix(data-table): replace hardcoded 100px min width with projected width

emoralesb05:
Also add possibility to override projected min width via configuration

closes https://github.com/Teradata/covalent/issues/891

### What's included?
- `ng serve`
- See that now the projected min width per column isnt 100px, but whatever the column header width is at a minimum
- Override it via column configuration

#### Test Steps

- [ ] do this
- [ ] then this
- [ ] finally this

- [ ] `npm run lint` passes.

### DIFF
--- a/src/platform/core/data-table/data-table.component.ts
+++ b/src/platform/core/data-table/data-table.component.ts
@@ -15,7 +15,7 @@ import { Subject } from 'rxjs/Subject';
 import { debounceTime } from 'rxjs/operator/debounceTime';
 
 import { TdDataTableRowComponent } from './data-table-row/data-table-row.component';
-import { ITdDataTableSortChangeEvent } from './data-table-column/data-table-column.component';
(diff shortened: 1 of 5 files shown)


src/platform/core/data-table/data-table.component.ts
```diff
@@ -968,9 +958,11 @@ export class TdDataTableComponent implements ControlValueAccessor, OnInit, After
         this._widths[index].limit = true;
       }
     }
-    // if there wasnt any width provided, we set a min of 100px
-    if (this._widths[index].value < 100) {
-      this._widths[index].value = 100;
+    // if there wasnt any width or min width provided, we set a min to what the column width min should be
+    if (!this._widths[index].limit &&
```
jeremysmartt: remove the this._widths[index].limit and just trigger off this.columns[index].width

PR #898, opened 2017-10-01
feat(data-table): allow dynamic row heights in data table even when using its virtual scroll

emoralesb05:
## Description

now we will cache the height of every row and recalculate the height as we keep scrolling

#### Test Steps

- [ ] `ng serve`
- [ ] Go to http://localhost:4200/#/components/data-table
- [ ] See new demos and play around with them

- [ ] `npm run lint` passes.

### DIFF
--- a/src/platform/core/data-table/data-table.component.ts
+++ b/src/platform/core/data-table/data-table.component.ts
@@ -73,6 +73,8 @@ export interface IInternalColumnWidth {
   max?: boolean;
 }
 
+const TD_VIRTUAL_OFFSET: number = 2;
+
 @Component({
   providers: [ TD_DATA_TABLE_CONTROL_VALUE_ACCESSOR ],
   selector: 'td-data-table',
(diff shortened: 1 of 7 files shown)


src/platform/core/data-table/data-table.component.ts
```diff
@@ -73,6 +73,8 @@ export interface IInternalColumnWidth {
   max?: boolean;
 }
 
+const TD_VIRTUAL_OFFSET: number = 2;
```
jeremysmartt: add comments of what this is

src/platform/core/data-table/data-table.component.ts
```diff
@@ -103,74 +105,46 @@ export class TdDataTableComponent implements ControlValueAccessor, OnInit, After
...
-   * For now we assume thats 49px.
-   */
-  get rowHeight(): number {
-    return 49;
-  }
+  private _rowHeightCache: number[] = [];
+  private _totalHeight: number = 0;
+  private _hostHeight: number = 0;
+  private _scrollVerticalOffset: number = 0;
+  private _offsetTransform: SafeStyle;
```
jeremysmartt: add comments on what all these are for

src/platform/core/data-table/data-table.component.ts
```diff
@@ -182,6 +156,7 @@ export class TdDataTableComponent implements ControlValueAccessor, OnInit, After
 
   /** internal attributes */
   private _data: any[];
+  private _virtualData: any[];
```
jeremysmartt: add comments on what this is

src/platform/core/data-table/data-table.component.ts
```diff
@@ -434,8 +414,14 @@ export class TdDataTableComponent implements ControlValueAccessor, OnInit, After
    */
   ngOnInit(): void {
     // initialize observable for resize calculations
-    this._resizeSubs = debounceTime.call(this._onResize.asObservable(), 10).subscribe(() => {
+    this._resizeSubs = this._onResize.asObservable().subscribe(() => {
+      if (this._rows) {
+        this._rows.toArray().forEach((row: TdDataTableRowComponent, index: number) => {
+          this._rowHeightCache[this.fromRow + index] = row.height + 1;
```
jeremysmartt: instead of caching everything just cache the ones you are looking at.  You can recalculate as needed when you scroll up and down.  With only a few rows shown should be quick.

src/platform/core/data-table/data-table.component.ts
```diff
@@ -983,4 +957,68 @@ export class TdDataTableComponent implements ControlValueAccessor, OnInit, After
...
+  private _calculateVirtualRows(): void {
+    let scrolledRows: number = 0;
+    if (this._data) {
+      this._totalHeight = 0;
+      let rowHeightSum: number = 0;
+      // loop through all rows to see if we have their height cached
+      // and sum them all to calculate the total height
+      this._data.forEach((d: any, index: number) => {
+        if (!this._rowHeightCache[index]) {
+          this._rowHeightCache[index] = this._rowHeightCache[0] || 48;
```
jeremysmartt: same as comment above about not caching everything

jeremysmartt: what is 48?  Maybe put that in a class level variable with a descriptive name and comments describing.

src/platform/core/data-table/data-table.component.ts
```diff
@@ -983,4 +957,68 @@ export class TdDataTableComponent implements ControlValueAccessor, OnInit, After
...
+        toRow = this._data.length;
+      } else if (!isFinite(toRow)) {
+        toRow = TD_VIRTUAL_OFFSET;
+      }
+      this._toRow = toRow;
+    } else {
+      this._totalHeight = 0;
+      this._fromRow = 0;
+      this._toRow = 0;
+    }
```
jeremysmartt: lots of calculation going on here, add some comments of what it all is doing.

src/platform/core/data-table/data-table.component.ts
emoralesb05: This actually just loops through only the ones rendered by the component.

emoralesb05: The thing we can improve later on is to modify the `totalHeight` here instead of recalculating it every time.

PR #905, opened 2017-10-03
fix(file-upload): clear internal file input on cancel

quanterion:
## Description
Currently if the user selects a file in the file upload, then presses cancel button and selects the same file again (select) event doesn't emit because in internal file-input value stays the same.

## What's included
This fix clears internal TdFileInputComponent inside TdFileUploadComponent.cancel() method.


emoralesb05:
Awesome, im trying to test this with and without the fix.. but i cant seem to replicate it. Am i missing something?

quanterion:
Here is a plunk that illustrates the problem http://plnkr.co/edit/P5kevR?p=preview
You should click two times on Upload file button and select the same file both times. File select event fires only once

emoralesb05:
Awesome man, yeah just saw! thanks for the PR

PR #935, opened 2017-10-18
Prioritize column setting for sortable

### LINKED ISSUE #933: Feature: DataTable sortable prioritize column
#### Feature Request
Change the behavior of sortable in the data-table.

Currently if the _general_ sortable boolean on the `td-data-table` is set to true it affects all columns.
I would recommend to allow an override on column basis.
e.g. Enable sortable on all columns, Disable sortable on specific column

In my personal opinion the settings on column basis should be weight higher than the _general_ setting.
Therefore it would be nice to be able to override a setting on column basis.

Affects all versions

##### Other information
I did a small fix in the following commit which solves this issue:
https://github.com/chdanielmueller/covalent/commit/42e36cd5decfe7ba679cf4c80c2060bc75be088b

In compliance with the contribution guidelines I will wait with opening a PR until this issue is approved.

chdanielmueller:
## Description
Change the behavior of sortable in the data-table as described in #933
closes #933

### What's included?
- Prioritize per column setting of sortable (allow override)

#### Test Steps
- [x] Set sortable on data table to true
- [x] Set sortable on one column to false
- [x] Check if sortable is disabled on the one specific column

- [x] `npm run lint` passes.

### DIFF
--- a/src/app/components/components/data-table/data-table.component.html
+++ b/src/app/components/components/data-table/data-table.component.html
@@ -274,7 +274,7 @@
             export class Demo {
               columns: ITdDataTableColumn[] = [
                 { name: 'first_name',  label: 'First Name', sortable: true, width: 150 },
-                { name: 'last_name', label: 'Last Name', filter: true },
+                { name: 'last_name', label: 'Last Name', filter: true, sortable: false },
                 { name: 'gender', label: 'Gender', hidden: false },
                 { name: 'email', label: 'Email', sortable: true, width: 250 },
                 { name: 'balance', label: 'Balance', numeric: true, format: DECIMAL_FORMAT },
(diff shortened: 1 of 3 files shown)


emoralesb05:
Thanks for the PR! Can you add in the demo 1 or 2 columns with `sortable:false` to show case it? 

here 
https://github.com/Teradata/covalent/blob/develop/src/app/components/components/data-table/data-table.component.html#L275-L281

and here 

https://github.com/Teradata/covalent/blob/develop/src/app/components/components/data-table/data-table.component.ts#L92-L98

chdanielmueller:
@emoralesb05 I disabled sorting on the last_name column in the demo.
I don't think any more explanation would be needed for this behavior.
What do you think?

emoralesb05:
Yup, agreed. I think thats fine to show case it. Thanks!

PR #942, opened 2017-10-21
chore(data-table): td-data-table missing (rowClick) at "Code" tab

dcarrask:
There is not call to action when a row is clicked at the documentation.

## Description
Just completed the documentation. Included at the td-data-table element the call to action placed at the Code Tab.


emoralesb05:
Thanks for that, if we are gonna add the `output` then we need to add the `showAlert()` method in the `TS` demo also.

dcarrask:
added the showAlert() method at TS documentation

Issue #727, opened 2017-07-01
td-file-upload Required Missing


ranouf:
Hi,

I would like to force the user to upload at least one file and sometime more.

There is no required field, do you know a solution to solve my problem?

emoralesb05:
Yeah, we need to add `ngModel` support to both `td-file-upload` and `td-file-input` so we can add `required` and such to them.

PR #951, opened 2017-10-30
feat(file-upload): required support added

richavyas:
## Description
This PR adds `ngModel` support so that `required` can be used in the usage template to enforce at least one file to be uploaded. Takes care of #727 

### What's included?
- ngModel support
- require support
- Control Value Accessors in the core file upload component

#### Test Steps
- [ ] Navigate to http://localhost:4200/#/components/file-upload
- [ ] check the required star shown on the Choose a file button in case `required` is added in the <td-file-upload> template 

- [ ] `npm run lint` passes.


src/app/components/components/file-upload/file-upload.component.html
```diff
@@ -9,10 +9,14 @@
...
         <p>Select Event: {{fileSelectMsg}}</p>
         <p>Upload Event: {{fileUploadMsg}}</p>
-        <td-file-upload #singleFileUpload (select)="selectEvent($event)" (upload)="uploadEvent($event)" (cancel)="cancelEvent()" [disabled]="disabled">
+        <td-file-upload #singleFileUpload [(ngModel)]="files" required (select)="selectEvent($event)" (upload)="uploadEvent($event)" (cancel)="cancelEvent()" [disabled]="disabled">
           <mat-icon>file_upload</mat-icon><span>{{ singleFileUpload.files?.name }}</span>
           <ng-template td-file-input-label>
-            <mat-icon>attach_file</mat-icon><span>Choose a file...</span>
+            <mat-icon>attach_file</mat-icon>
+            <span>
+              Choose a file...
```
kyleledbetter: can we abstract this out of the code so it's easy for the user to change and translate?

kyleledbetter: ohh derp it's in a template

emoralesb05:
Pushed a few commits to fix the `ngModel` support since it wasnt working properly

PR #1007, opened 2017-12-04
feat(data-table): add index to row select and row click outputs

### LINKED ISSUE #992: td-data-table (rowSelect) get selected index
#### Feature Request
Could the td-data-table return an object with selected index ontop of selected row data

StackBlitz template: https://stackblitz.com/edit/covalent-data-tables

Currently it returns
```json
{"row":row_data,"selected":true}"
```
I would expect an object with form
```json
{"row":{row_data},"index":3,"selected":true}"
```
Returning the row index eases filtering of selected rows especially on deselecting the row to remove it from the selectedRows array is easier since we know its index

emoralesb05:
## Description
`index` property is added to both `rowClick` and `rowSelect` output events.

- [ ] `npm run lint` passes.

closes #992

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "d3": "^4.4.0",
     "hammerjs": "^2.0.8",
     "highlight.js": "9.11.0",
-    "rxjs": "^5.5.2",
+    "rxjs": "5.5.2",
     "showdown": "1.6.4",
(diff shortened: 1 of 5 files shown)


nueko:
hey, how to get index from `tdDataTableTemplate`?

PR #1019, opened 2017-12-08
feat(): use secondary entry points for @angular/material and rxjs

### LINKED ISSUE #1001: Unused @angular/material scripts being pulled to application bundle
#### Bug Report
I just include the `TdSteps` module in the application. However, I see several other parts of @angular/material, such as chips, autocomplete, snackbar, etc.. being pulled into the generated code.

##### Screenshots
Below is the output for source-map-explorer when I don't import the `TdStepsModule`:
image

And here is the output after importing the `TdStepsModule`:
image

No additional material code should be included in the bundle

Reduce the application size, which improves the user experience

Angular 5.0.3
Material 5.0.0-rc.2
Teradata covalent 1.0.0-rc.0

Not tested

##### Other information
I tried to chase which covalent script was causing the issue, but couldn't find it

emoralesb05:
we need to cherry pick the modules we use as dependencies in covalent so we dont implicitly load everything that comes in those packages when we only need a few things.. thus reducing bundle sizes.

this is needed so we dont load all the @angular/material packages when we only need a few (same for rxjs)

e.g.
```
import { MatCardModule, MatInputModule } from '@angular/material';
import { map, startsWith } from 'rxjs/operators';
```
would be
```
import { MatCardModule } from '@angular/material/card';
import { MatInputModule } from '@angular/material/input';

import { map } from 'rxjs/operators/map';
import { startsWith } from 'rxjs/operators/startsWith';
```

### What's included?
- Lint rule so we cant import directly from `rxjs`, `rxjs/operators`, `@angular/cdk` and `@angular/material`
- Cherry pick only the modules we need

#### Test Steps
- [ ] `npm run serve`
- [ ] See everything working the same way

- [ ] `npm run lint` passes.

closes #1001

### DIFF
--- a/scripts/rollup.js
+++ b/scripts/rollup.js
@@ -34,7 +34,24 @@ gulp.task('rollup-code', '', function() {
     '@angular/platform-browser': 'ng.platformBrowser',
     '@angular/platform-browser/animations': 'ng.platformBrowser.animations',
     '@angular/platform-browser-dynamic': 'ng.platformBrowserDynamic',
+    // Material entry points
     '@angular/material': 'ng.material',
+    '@angular/material/core': 'ng.material.core',
+    '@angular/material/input': 'ng.material.input',
+    '@angular/material/button': 'ng.material.button',
+    '@angular/material/sidenav': 'ng.material.sidenav',
+    '@angular/material/autocomplete': 'ng.material.autocomplete',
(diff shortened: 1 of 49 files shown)


aciccarello:
Related to angular/material2#6998

Issue #278, opened 2017-01-22
Lint for css/scss


emoralesb05:
#### Feature Request

Possibly use https://github.com/stylelint/stylelint to lint css so we can standardize it too.

emoralesb05:
https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#configuration

PR #1052, opened 2017-12-20
chore(tooling): add stylelint to start linting scss and initial rule set

emoralesb05:
the rule set is just a draft, the rules are subject to change.

### What's included?

- `npm run stylelint` to do initial linting for scss files
- Rename `npm run lint` to `npm run tslint`

#### Test Steps

- [ ]  Install VS Code extension `stylelint`
- [ ] `npm run stylelint` and see all the errors (we will fix in consecutive PR's)
- [ ] `npm run tslint` and see it work

closes https://github.com/Teradata/covalent/issues/278


jotatoledo:
You could consider to add the markdown/javascript linters as well

emoralesb05:
@jotatoledo any suggestions on a linter for html?? would be nice we can add custom rules for angular attrs also

jotatoledo:
Havent found any yet ;(

package.json
```diff
@@ -10,7 +10,8 @@
   ],
   "scripts": {
     "webdriver-update": "bash ./node_modules/.bin/webdriver-manager update",
-    "lint": "tslint --format codeFrame -c ./tslint.json \"./src/**/*.ts\" -e \"./src/**/typings.d.ts\" -e \"./src/environments/**\"",
+    "tslint": "./node_modules/.bin/tslint --format codeFrame -c ./tslint.json \"./src/**/*.ts\" -e \"./src/**/typings.d.ts\" -e \"./src/environments/**\"",
+    "stylelint": "./node_modules/.bin/stylelint 'src/**/*.scss' '!**/_layout.scss' '!**/markdown.component.scss' --config .stylelintrc.json --syntax scss",
```
jeremysmartt: should we also have a target that does both tslint and stylelint in command?

emoralesb05: I was debating that also, i think we can use `npm run lint` to run both

jeremysmartt: 👍

PR #1061, opened 2018-01-03
feat(chips): add `inputPosition` input to change its position to before or after (closes #990)

### LINKED ISSUE #990: make td-chips and stacked more flexible using containers
#### Feature Request
if we could make the td-chips and stacked more flexible using containers for the stacked chips and for the input itself
#### What is the expected behaviour?
when using the stacked td-chips I was struggling to make the make:
1. input to be placed above the stacked list (and not below of the stacked list)
2. stacked to be in a container/div to allow scroll on it once it has too many rows (as my form has a limit view in screen).

#### What is the motivation / use case for changing the behaviour?
just to make this component working with stack as flexible as possible

emoralesb05:
## Description
To allow a bit more flexibility, we added a new input to change where the input is going to be positioned.. whether is before or after the list of chips regardless if its stacked or not.

### What's included?

- `[inputPosition]` input with `before` and `after` values

#### Test Steps

- [ ] `npm run serve`
- [ ] Go to chips demo
- [ ] See before option change

image

image

closes https://github.com/Teradata/covalent/issues/990


emoralesb05:
Fixed the build and linter. So everything should be fine now.

PR #1072, opened 2018-01-05
chore(deps): update to node 8.9.4 + pa11y support and CI test

ilsiepotamus:
## Description
Now that Angular supports node 8, Covalent should also support node 8.  Using the LTS version, `carbon`, currently at 8.9.4.

### What's included?
- Dependency upgrades for node 8 (specifically, `require-dir`)
- pa11y support (a11y testing) and addition to travis

#### Test Steps
- [ ] `npm run a11y` works.


scripts/a11y/a11y.sh
```diff
@@ -0,0 +1,67 @@
...
+    ;;
+  esac
+done
+
+shift $((OPTIND-1))
+
+[ "$1" = "--" ] && shift
+
+# run install.
+npm run reinstall
```
emoralesb05: Why do you need to do this here? 

Travis does this before running this so it should be fine.

ilsiepotamus: in case we migrate to something that's not travis.

ilsiepotamus: (plus, travis does `npm install`, not `npm run reinstall`)

PR #1122, opened 2018-02-14
fix(docs): replace localhost

ekoeryanto:
### What's included?

- One
- Two
- Three

#### Test Steps

- [ ] `npm run serve`
- [ ] then this
- [ ] finally this


emoralesb05:
Can you make this PR point to `develop` please?

Issue #1133, opened 2018-03-28
Angular 6 & rxjs 6


Enngage:
Hey,

Any chance you can start preparing for Angular 6 with rxjs 6? I've recently updated my project to rxjs 6 only to found out that I can no longer compile the solution:

![aa](https://user-images.githubusercontent.com/7137351/38022115-f3922848-327e-11e8-9997-eb8b8bf615ef.png)

This happens also with the [rxjs-compat ](https://www.npmjs.com/package/rxjs-compat) package installed.

Edit: The workaround I managed to get it working was  `node_modules\@covalent\core\esm5\covalent-core.js` 

```javascript
import { toPromise } from 'rxjs/operators/toPromise';
```
to

```javascript
import 'rxjs/add/operator/toPromise';
```

This seems like an issue with rxjs library which I raised in their repo.

emoralesb05:
Yup, we are gonna start again with the `beta` process getting ready for `angular 6`.. so expect this soon 😄

Enngage:
Awesome, happy to hear that :-)

emoralesb05:
Just FYI, started the process here https://github.com/Teradata/covalent/pull/1136 but we still have to test and fix a few things 😄

PR #1136, opened 2018-04-05
feat(): add support for angular@6

emoralesb05:
## Description
Now that angular is in RC mode. We will add support for `angular/material/rxjs 6`. 
This means a breaking change since covalent wont work with any version lower than 6.. so we will have to start the `2.0.0` beta process and take advantage of this to include new things and change some things for the better.

This doesnt mean 1.0.0 is dead yet, we will still support it and apply bugfixes to it until `2.0.0` is stable. But all new features will go directly to 2.0.0.

### What's included?

- Upgrade angular, material and rxjs to 6.0.0
- Apply breaking changes from them
- Update typescript to 2.7.2
- Update `@ngx-translate/core` and other dependencies that needed angular 6 support.
- Update `@angular/cli` to 1.7.3

#### Test Steps

- [x] `yarn install`
- [x] `npm run serve`
- [ ] See that everything still works the same as http://getcovalent.com

closes https://github.com/Teradata/covalent/issues/1133


aciccarello:
Related https://github.com/mgechev/rxjs-tslint

PR #1155, opened 2018-05-18
fix(dynamic-forms): removed regex for name

JulieMarie:
## Description
Removed regex from name string

### Test Steps
- [ ] checkout branch
- [ ] `npm run serve`

### General Tests for Every PR

##### Stackblitz with regex implemented
https://stackblitz.com/edit/ajc-stackoverflow-48705028?file=app/app.component.ts
covalent-fix


src/platform/dynamic-forms/services/dynamic-forms.service.ts
```diff
@@ -49,7 +49,7 @@ export interface ITdDynamicElementConfig {
   validators?: ITdDynamicElementValidator[];
 }
 
-export const DYNAMIC_ELEMENT_NAME_REGEX: RegExp = /^[a-zA-Z]+[a-zA-Z0-9-_]*$/;
+export const DYNAMIC_ELEMENT_NAME_REGEX: RegExp = /^[^0-9][^\@]*$/;
```
ilsiepotamus: what are you trying to capture with this regex?

JulieMarie: In the original unit tests, it's testing for a name starting with a number or a name with an ampersand in it.  I do realize that the `\` isn't necessary to escape the `@`.

ilsiepotamus: ok, so it actually is testing for what you want to be testing for, however.  I just thought it was an odd pattern.

JulieMarie: Agreed.

PR #1182, opened 2018-06-20
chore(): modify npm script and nightly script for experimental package

stevenov7:
## Description

After speaking with Ed + Jeremy, experimental module should be bound to Covalent's versioning rather then be on it's own versioning and it should be consumed though it's own nightly build.

### What's included?

- Bound experimental's package.json `version` property to Covalent's build process.
- Experimental nightly build integration

#### Test Steps

- [ ] `yarn build:lib`
- [ ] Verify `/deploy/platform/experimental/package.json` `version` matches the other entry points `version` property (EX: core and experimental `version` are both 2.0.0-beta-2 after the build)
// TODO: update with night build test instructions 

### DIFF
--- a/docs/DEVELOPER_GUIDE.md
+++ b/docs/DEVELOPER_GUIDE.md
@@ -15,7 +15,7 @@ The source has two parts, the platform and a documentation app.
 
 ## Adding a new platform feature (component/module/utility)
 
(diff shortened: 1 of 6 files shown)


stevenov7:
Going to be adding updates to this PR that will make the experimental module consumable only though a nightly build. This feedback is based on communication between @emoralesb05 and @jeremysmartt during the Covalent Standup.

PR #1183, opened 2018-06-20
feat(breadcrumbs): Initial scaffolding for breadcrumbs

jeremysmartt:
## Description
In work, Don't merge

### What's included?

- One
- Two
- Three

#### Test Steps

- [ ] `npm run serve`
- [ ] then this
- [ ] finally this

### DIFF
--- a/src/platform/experimental/breadcrumbs/README.md
+++ b/src/platform/experimental/breadcrumbs/README.md
@@ -1,0 +1,14 @@
+// TODO: add respective info about your component here
+# TdBreadcrumbsComponent
+
+**NOTE:** 
+
+## API Summary
+
+#### Inputs
+
+#### Methods
+
+## Setup
+
+## Usage
--- a/src/platform/experimental/breadcrumbs/breadcrumbs.component.ts
+++ b/src/platform/experimental/breadcrumbs/breadcrumbs.component.ts
@@ -1,0 +1,14 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'td-breadcrumbs',
+  styleUrls: ['./breadcrumbs.component.scss'],
+  templateUrl: './breadcrumbs.component.html',
+})
+export class TdBreadcrumbsComponent {
+
+  constructor() {
+    // stub
+  }
+
+}
--- a/src/platform/experimental/breadcrumbs/breadcrumbs.component.html
+++ b/src/platform/experimental/breadcrumbs/breadcrumbs.component.html
@@ -1,0 +1,1 @@
+Breadcrumbs here
--- a/src/platform/experimental/breadcrumbs/breadcrumbs.module.ts
+++ b/src/platform/experimental/breadcrumbs/breadcrumbs.module.ts
@@ -1,0 +1,19 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+
+import { TdBreadcrumbsComponent } from './breadcrumbs.component';
+
+@NgModule({
+  imports: [
+    CommonModule,
+  ],
+  declarations: [
+    TdBreadcrumbsComponent,
+  ],
+  exports: [
+    TdBreadcrumbsComponent,
+  ],
+})
+export class CovalentBreadcrumbsModule {
+
+}
--- a/src/platform/experimental/breadcrumbs/index.ts
+++ b/src/platform/experimental/breadcrumbs/index.ts
@@ -1,0 +1,1 @@
+export * from './public-api';
--- a/src/platform/experimental/breadcrumbs/package.json
+++ b/src/platform/experimental/breadcrumbs/package.json
@@ -1,0 +1,7 @@
+{
+  "ngPackage": {
+    "lib": {
+      "entryFile": "index.ts"
+    }
+  }
+}
--- a/src/platform/experimental/breadcrumbs/public-api.ts
+++ b/src/platform/experimental/breadcrumbs/public-api.ts
@@ -1,0 +1,2 @@
+export * from './breadcrumbs.module';
+export * from './breadcrumbs.component';
--- a/src/platform/experimental/public-api.ts
+++ b/src/platform/experimental/public-api.ts
@@ -1,1 +1,2 @@
 export * from './template-rename-me-experiment-module/index';
+export * from './breadcrumbs/index';


src/platform/experimental/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  /**
+   * Sets the icon url shown between breadcrumbs. Defaults to right chevron.
+   */
+  @Input('separatorIcon') separatorIcon: string = 'navigate_next';
+
+  constructor(private _elementRef: ElementRef, private _changeDetectorRef: ChangeDetectorRef) { }
+
+  ngAfterViewInit(): void {
+    this._resizeSubscription = merge(
```
jotatoledo: Shouldnt this subscription be cleaned up on component destroy?

jeremysmartt: @jotatoledo You are correct, nice catch.

src/platform/experimental/breadcrumbs/breadcrumbs.component.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [
+        TdBreadcrumbsTestComponent,
+        FakeComponent,
+      ],
+      imports: [
+        NoopAnimationsModule,
+        RouterTestingModule.withRoutes([
```
emoralesb05: Is there any reason to use mock routes?

src/platform/experimental/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  private displayWidthAvailableCrumbs(): void {
+    let curTotCrumbWidth: number = 0;
+    let crumbsArray: TdBreadcrumbComponent[] = this._breadcrumbs.toArray();
+    // calculate the current width of the shown breadcrumbs
+    for (let i: number = this.hiddenBreadcrumbs.length; i < crumbsArray.length; i++) {
+      curTotCrumbWidth += crumbsArray[i].width;
+    }
+    let winWidth: number = this._elementRef.nativeElement.parentElement.offsetWidth - this._windowWidthPadding;
```
emoralesb05: Should check the host here instead of the parent in case the `component width` changes.

src/platform/experimental/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,127 @@
...
+  styleUrls: ['./breadcrumbs.component.scss'],
+  templateUrl: './breadcrumbs.component.html',
+})
+export class TdBreadcrumbsComponent implements AfterViewInit, DoCheck, AfterContentChecked {
+
+  private _resizeSubscription: Subscription;
+  private _widthSubject: Subject<number> = new Subject<number>();
+  private _resizing: boolean = false;
+  // Padding on the window when calculating the full size
+  private _windowWidthPadding: number = 20;
```
emoralesb05: Is this needed if we check the host width rather than the parent width?

src/platform/experimental/breadcrumbs/README.md
```diff
@@ -0,0 +1,101 @@
...
+  <a td-breadcrumb [routerLink]="'/layouts2'">Layouts2</a>
+  <a td-breadcrumb [routerLink]="'/layouts3'">Layouts3</a>
+  <td-breadcrumb class="tc-grey-500">Manage List</td-breadcrumb>
+</td-breadcrumbs>
+```
+
+Example with all inputs/outputs: 
+
+```html
+<div layout-gt-sm="row" tdMediaToggle="gt-xs" [mediaClasses]="['push-sm']">
```
emoralesb05: Lets remove this div and just keep the example with the component usage.

src/platform/experimental/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,127 @@
...
+          this._resizing = false;
+          this._changeDetectorRef.markForCheck();
+        }, 100);
+      }
+    });
+  }
+
+  ngDoCheck(): void {
+    if (this._elementRef && this._elementRef.nativeElement) {
+      this._widthSubject.next((<HTMLElement>this._elementRef.nativeElement).getBoundingClientRect().width);
```
jotatoledo: What about hiding the argument of `next` behind a private getter?
Something like private `get nativeElementWidth(): number {}`

src/platform/experimental/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,127 @@
...
+   * Set prebaked styles on the crumbs depending on if 
+   * they are the first or last crumb
+   */
+  private setStyles(): void {
+    let breadcrumbArray: TdBreadcrumbComponent[] = this._breadcrumbs.toArray();
+    if (breadcrumbArray && breadcrumbArray.length > 0) {
+      // don't show the right chevron on the first breadcrumb
+      breadcrumbArray[0].displayIcon = false;
+    }
+    breadcrumbArray.map((breadcrumb: TdBreadcrumbComponent) => {
```
jotatoledo: Wouldnt a `forEach` be better suited for iterating and mutating the elements in the array? considering that you arent really mapping the array elements into something new.

src/platform/experimental/breadcrumbs/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+# td-breadcrumb
+
+`td-breadcrumb` element generates an individual breadcrumb component.
+
+## API Summary
+
+#### Methods
+
++ displayIcon: function(): boolean
```
emoralesb05: Not sure if we state that these properties are public. Since we use them internally to change the state of it.. which we should probably add a `_` as a prefix, since that is the standard for `internal` properties.

src/platform/experimental/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,127 @@
...
+    return this._breadcrumbs ? this._breadcrumbs.length : 0;
+  }
+
+  /**
+   * Set prebaked styles on the crumbs depending on if 
+   * they are the first or last crumb
+   */
+  private setStyles(): void {
+    let breadcrumbArray: TdBreadcrumbComponent[] = this._breadcrumbs.toArray();
+    if (breadcrumbArray && breadcrumbArray.length > 0) {
```
jotatoledo: Im almost sure that there is no need to check the existence of `breadcrumbArray`.
The not empty conditional could be shortened to `if(breadcrumbArray.length)`

src/platform/experimental/breadcrumbs/breadcrumbs.component.spec.ts
jeremysmartt: just so the test doesn't fail with the routherlink in there

src/platform/experimental/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }
+
+  ngAfterContentInit(): void {
+    this.setCrumbIcons();
+    this._changeDetectorRef.markForCheck();
+  }
+
+  ngOnDestroy(): void {
+    if (this._resizeSubscription) {
```
jotatoledo: Avoid the truthy check by declaring the sub variable with the initial value of `Subscription.EMPTY`;

emoralesb05: Huh, i didnt know about that.

jeremysmartt: 👍

emoralesb05:
Just noticed that the icon is also clickable.. wonder if that will cause issues 🤔

emoralesb05:
- [x] pushed changed to make sure clicking on the icon doesnt trigger navigation

PR #1199, opened 2018-07-18
feat(): install covalent tools as dependency

emoralesb05:
## Description

this is to leverage the global tslint and stylelint files across
all our product portfolio

### What's included?

- added @covalent/tools as dependency
- point to covalent tools tslint file
- point to covalent tools stylelint file

#### Test Steps

- [x] `yarn install`
- [x] `npm run tslint` should work the same way (try forcing linting errors)
- [x] `npm run stylelint` should work the same way (try forcing linting errors)
- [x] `./node_modules/.bin/cov versions` should work


.stylelintrc.json
```diff
@@ -1,96 +1,3 @@
...
-    "selector-max-id": 0,
-    "no-missing-end-of-source-newline": true,
-    "no-duplicate-selectors": true,
-    "no-extra-semicolons": true,
-    "no-invalid-double-slash-comments": true,
-    "max-line-length": 120,
-    "max-empty-lines": 1,
-    "media-feature-name-no-unknown": true
-  }
+  "extends": "./node_modules/@covalent/tools/lint/stylelint/.stylelintrc.json"
```
jeremysmartt: awesome

PR #1201, opened 2018-07-21
feat(dynamic-forms): add support for Materials multiselect form field (closes #820)

### LINKED ISSUE #820: Feature request: Dynamic forms checkboxes from array
There are many cases when I'd like the user to be able check a number of related boxes and have the results be turned into an array.

Ideal:

```
filters: ITdDynamicElementConfig[] = [{
    name: "userChoices",
    type: TdDynamicType.CheckboxArray,
    selections: [
        {name: 'choice1', label: 'Choice One'}
        {name: 'choice2', label: 'Choice Two'}
        {name: 'choice3', label: 'Choice Three'}
    ],
    default: ['choice1', 'choice3'],
    required: true //this would throw an error if nothing is selected
    }]
```

This would display something like this:

- [x] Choice One
- [ ] Choice Two
- [x] Choice Three

And provide an array on the backend like:

```
['choice1','choice3']
```

This is particularly useful when assembling filters where a user may be selecting many options and it's not convenient to manually check the boolean state of every checkbox as the developer.

markonyango:
## Description
I have added Materials multiselect capability to the already existing TdDynamicElement.Select to allow developers to provide Materials select form field, containing multiple selection options that can be selected at the same time ([Material Docs](https://material.angular.io/components/select/overview#multiple-selection))

#### Test Steps
-  `yarn run serve:experimental`
- open localhost:4200 in your browser
- navigate to the multiselect section to see a variety of implementations of the new select field (for futher implementation variety details, check the multiselect folder in the test-bed folder)

closes https://github.com/Teradata/covalent/issues/820


markonyango:
I overlooked the type definition in dynamic-select.component.ts . It being any[] does not allow objects to be passed to selections. Setting it to any would be a viable solution but not the best. Once I found the most elegant way to change this, I am pulling this PR myself.

src/platform/dynamic-forms/dynamic-element.component.ts
```diff
@@ -100,6 +100,11 @@ export class TdDynamicElementComponent extends AbstractControlValueAccessor
    */
   @Input() selections: any[] = undefined;
 
+  /**
+   * Sets multiple property for array elements (if supported by element).
+   */
+  @Input() multi: boolean = undefined;
```
emoralesb05: My only feedback with this would be to name it `multiple` so it is more inline with HTML5 standards.

Everything else looks good.

markonyango: I agree with you on standards and thank you for your quick feedback.

emoralesb05:
Thanks for the contribution!

PR #1209, opened 2018-08-11
fix(breadcrumbs): clean up imports of observable creation methods

### LINKED ISSUE #1208: fix(breadcrumbs): import rxjs creation methods from 'rxjs'
#### Bug Report

The current breadcrumbs component implementation does not import `merge` and `fromEvent` from 'rxjs', which is not the correct way for rxjs 6.x and thus fails on build if `rxjs-compat` isnt installed.

Both creation methods should be imported from 'rxjs'
##### Other information

See it in [the source code](https://github.com/Teradata/covalent/blob/develop/src/platform/experimental/breadcrumbs/breadcrumbs.component.ts#L23)

jotatoledo:
* Change to import creation methods from rxjs. Without this change, consumers
of covalent are obligated to install rxjs-compat.

Closes #1208

### What's included?
Minor fix to imports in an experimental module.

#### Test Steps
Does not apply.

- [X] `npm run tslint` passes.
- [X] `npm run stylelint` passes.

### DIFF
--- a/src/platform/experimental/breadcrumbs/breadcrumbs.component.ts
+++ b/src/platform/experimental/breadcrumbs/breadcrumbs.component.ts
@@ -15,13 +15,13 @@ import {
 import {
   Subscription,
   Subject,
+  fromEvent,
+  merge,
 } from 'rxjs';
 import {
   debounceTime,
   distinctUntilChanged,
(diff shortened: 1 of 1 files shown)


jeremysmartt:
Thank you for the PR 👍

Issue #1126, opened 2018-02-21
Add disabled option to dynamic-form select-component


benjaminleetmaa:
#### Feature Request
Add disabled option to dynamic-form select-component

aciccarello:
Related to #490

emoralesb05:
#1215 will take care of this

PR #1215, opened 2018-08-13
 feat(dynamic-forms): allow elements to be disabled via configuration

emoralesb05:
## Description

adding a disabled property to disable properties on the fly without the
need of accessing the underlying form

### What's included?

- new disabled property in the configuration
- updated Demo and README

#### Test Steps

- [ ] `npm run serve`
- [ ] Go to http://localhost:4200/#/components/dynamic-forms
- [ ] Play with form build to see how the disable state affects different components

closes https://github.com/Teradata/covalent/issues/1126


JulieMarie:
disabled-elements

I don't like the how the disabled slider looks/works.  It's not grayed out, but the dot grays out when you click on it.  Title should be grayed out too.

Don't like the slider toggle.  Title should be grayed out and larger font to match the others.

File upload button is still active even though element is disabled.

PR #1222, opened 2018-08-15
fix(dynamic-forms): make sure hints and errors dont overlap

emoralesb05:
## Description

properly leveraging the mat-error element on components that use
mat-form-field fixes this issue, but to achieve this we need to
refactor how the error message is shown

also add documentation on how to show the error message when creating a
custom component

### What's included?

- use mat-error when possible
- add info in README and demo on how to show error in custom component

#### Test Steps

- [ ] `npm run serve`
- [ ] Go to http://localhost:4200/#/components/dynamic-forms
- [ ] Check out how to set the error in a custom component in the custom components DEMO


JulieMarie:
LGTM

PR #1231, opened 2018-08-28
feat(notification): limit added

pv-arunadevi:
## Description
Limit attribute is added to the td-notification-count component. When the notification count exceeds the limit set, it would display as {limit}+. It was initially hard coded as 99. Now it defaults to 99 but can be overridden by the user of this component.

### What's included?

- limit attribute added

#### Test Steps

- [x] `npm run serve`

### DIFF
--- a/src/app/components/components/notifications/notifications.component.html
+++ b/src/app/components/components/notifications/notifications.component.html
@@ -20,7 +20,7 @@
             <mat-icon>messages</mat-icon>
           </td-notification-count>
           <button class="overflow-visible" mat-icon-button [matMenuTriggerFor]="notificationsMenu">
-            <td-notification-count [notifications]="101">
+            <td-notification-count [notifications]="101" [limit]=50>
               <mat-icon>notifications</mat-icon>
             </td-notification-count>
           </button>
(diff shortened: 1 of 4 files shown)


kyleledbetter:
@pv-arunadevi awesome and thanks! Would you mind uploading an animated Gif using https://www.cockos.com/licecap/ and adding to your PR description just so we can visually confirm that the UI doesn't break with single, double or triple digits?

emoralesb05:
This looks great, from my perspective the things that are missing are possibly 2 things:

- Updated README.md of notifications to state that we have that input and how to use it.
- Maybe a demo in the documentation site using the input (more interested in the previous one).

pv-arunadevi:
screen shot 2018-08-28 at 4 03 35 pm
screen shot 2018-08-28 at 4 04 07 pm
screen shot 2018-08-28 at 4 04 50 pm
screen shot 2018-08-28 at 4 05 34 pm

pv-arunadevi:
README updated and screenshots with single, double and triple digits attached. Documentation site is updated.

emoralesb05:
Thanks for the contribution!

PR #1247, opened 2018-10-08
fix(chips): mobile select

### LINKED ISSUE #1044: [BUG] TdChips with Autocomplete in mobile does not select
#### Bug Report
Autocomplete options in mobile doesn't select.

Teradata issue screen shot giff

I'm using the latest release 1.0.0-rc.1

felipeik:
## Description
Fixed select on Chips Autocomplete on mobile devices.
Closes #1044

### What's included?
* Modified: `src/platform/core/chips/chips.component.ts`

* Reason of the bug: `touchend` events is canceling `click` events on mobile
* Fix explanation: Added debounceTime to prevent both events from being fired together

#### Test Steps
- [ ] Go to Chips demo page
- [ ] Try selecting autocomplete on desktop
- [ ] Try selecting autocomplete on mobile view
- [ ] With ng serve --host 0.0.0.0, try on a real phone opening the application on port 4200

### DIFF
--- a/src/platform/core/chips/chips.component.ts
+++ b/src/platform/core/chips/chips.component.ts
@@ -71,6 +71,7 @@ export class TdChipsComponent extends _TdChipsMixinBase implements IControlValue
   private _chipRemoval: boolean = true;
   private _focused: boolean = false;
   private _tabIndex: number = 0;
+  private _touchendDebounce: number = 100;
 
   _internalClick: boolean = false;
   _internalActivateOption: boolean = false;
@@ -722,6 +723,7 @@ export class TdChipsComponent extends _TdChipsMixinBase implements IControlValue
         fromEvent(this._document, 'click'),
         fromEvent(this._document, 'touchend'),
       ).pipe(
+        debounceTime(this._touchendDebounce),
         filter(
           (event: MouseEvent) => {
             const clickTarget: HTMLElement = <HTMLElement>event.target;


felipeik:
BTW, I removed the `touchend` event altogether and the application worked on a real iPhone/Chrome device.

I don't know if the `touchend` evnt is necessary for other OS or browsers.

src/platform/core/chips/chips.component.ts
```diff
@@ -722,6 +722,7 @@ export class TdChipsComponent extends _TdChipsMixinBase implements IControlValue
         fromEvent(this._document, 'click'),
         fromEvent(this._document, 'touchend'),
       ).pipe(
+        debounceTime(this.debounce),
```
emoralesb05: We probably should have a different constant (with 100-200 ms) for the debounce since this `debounce` variable is more about the key presses and opening the menu and its configurable, so it will be weird if they add a `1000` or `2000` debounce and it takes 1 - 2 seconds to select it.

emoralesb05:
Thanks for the investigative work! Just left a small comment 🎉

src/platform/core/chips/chips.component.ts
felipeik: just created a `this._touchendDebounce = 100` at the top of the file

felipeik:
Hi @emoralesb05 , I just pushed the fix to the comment!

PR #1248, opened 2018-10-08
fix(chips): long texts

### LINKED ISSUE #1102: chips bug not handling long text content well
bug

#### Bug Report
- Found this issue in AppCenter while working on restyling the "adding user flow/chips as an admin", put a temp fix atm into AppCenter need to remove it after this fix gets in.
- [ ] Go here https://teradata.github.io/covalent/#/components/chips
- [ ] Scroll down to the "Stacked chips and Autocomplete with custom inputs"
- [ ] Inspect the text within one of the chips (EX: "stepper" text chip)
- [ ] Enter a long string though the inspector, you should see that the chips looks odd (see gif below)

rec

- The chips should be able to handle long text along at the same time using `td-chip-avatar`

- Conflicting override styles in AppCenter (note: `span`):
```
// AppCenter styles.scss
// fix for long strings in chips
.td-chip {
  span {
    white-space: nowrap;
    text-overflow: ellipsis;
    overflow: hidden;
  }
}

//html
<td-chips ...>
        <ng-template td-chip let-chip="chip">
          <div class="tc-grey-100 bgc-teal-700" td-chip-avatar>{{chip.initials}}</div>
          <span class="mat-caption">{{ chip.nameInfo }}</span>
        </ng-template>
      </td-chips>
```
cause issues to the `td-chip-avatar` too look like its being cut in half issue appears to be coming from the these styles being applied to the internal <td-chips> `span`s. If we change in AppCenter the `span` css selector to be more specific like below. The issue does not occur but another issue arise with long text and the per chips "X" remove button.
 ```
// AppCenter styles.scss
// fix for long strings in chips
.td-chip {
  .chip-ellipsis {
    white-space: nowrap;
    text-overflow: ellipsis;
    overflow: hidden;
  }
}

//html
<td-chips ...>
        <ng-template td-chip let-chip="chip">
          <div class="tc-grey-100 bgc-teal-700" td-chip-avatar>{{chip.initials}}</div>
          <span class="chip-ellipsis mat-caption">{{ chip.nameInfo }}</span>
        </ng-template>
      </td-chips>
```

Covalent: 1.0.0-rc.2
Angular Material: 5.0.4
Chrome/Mac

No

### PR DESCRIPTION
**Pull request by felipeik**
## Description
Fixed long text inside chips
Closes #1102

### What's included?
**modified: src/platform/core/chips (multiple files modified)**

* Added `text-truncate` to chip content
* Added `min-width` to `.chip-content` to truncate flexbox children (see https://css-tricks.com/flexbox-truncated-text/)
* Simulated `flex="none"` to `.td-chip-content`

**modified: src/app/components/components/chips/chips.component.html (for truncate documentation)**
* Added `.text-truncate` class to docs

#### Test Steps
- [ ] Go to Chips page
- [ ] Inspect the text within one of the chips (EX: "stepper" text chip), it can be any chip
- [ ] Enter a long string inside `span` with class `.text-truncate` though the inspector
- [ ] Try the same with `chip-avatars` and close icons

- [ ] See the chip occupying 100% of the width of the parent container
- [ ] see the text is truncated 
- [ ] see the `chip-avatar` without distortion
- [ ] see the close icon still inside the chip

#### Before the fix:

bug

#### After the fix:

chip long text

### DIFF
--- a/src/platform/core/chips/chips.component.scss
+++ b/src/platform/core/chips/chips.component.scss
@@ -32,6 +32,8 @@
       align-items: center;
       align-content: center;
       justify-content: start;
+      // truncate content on flexbox children
+      min-width: 0;
       &.td-chip-stacked {
         // layout-align space-between center
         justify-content: space-between;
@@ -63,6 +65,9 @@
           width: 32px;
           @include rtl(margin, 0 8px 0 -12px, 0 -12px 0 8px);
           border-radius: 50%;
+          // emulate flex="none"
+          flex: 0 0 auto;
+          box-sizing: border-box;
         }
       }
       box-sizing: border-box;
--- a/src/app/components/components/chips/chips.component.html
+++ b/src/app/components/components/chips/chips.component.html
@@ -128,7 +128,8 @@
                   (inputChange)="filterObjects($event)"
                   requireMatch>
           <ng-template td-chip let-chip="chip">
-            <mat-icon>location_city</mat-icon> {{chip.city}}, (Pop: {{chip.population}})
+            <mat-icon>location_city</mat-icon>
+            <span class="text-truncate">{{chip.city}}, (Pop: {{chip.population}})</span>
           </ng-template>
           <ng-template td-autocomplete-option let-option="option">
             <div layout="row" layout-align="start center">
@@ -151,7 +152,8 @@
                       (inputChange)="filterObjects($event)"
                       requireMatch>
               <ng-template td-chip let-chip="chip">
-                <mat-icon>location_city</mat-icon> { {chip.city} }, (Pop: { {chip.population} })
+                <mat-icon>location_city</mat-icon>
+                <span class="text-truncate">{ {chip.city} }, (Pop: { {chip.population} })</span>
               </ng-template>
               <ng-template td-autocomplete-option let-option="option">
                 <div layout="row" layout-align="start center">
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/platform/core/chips/chips.component.scss [hunk 1]
```diff
@@ -63,6 +65,11 @@
           width: 32px;
           @include rtl(margin, 0 8px 0 -12px, 0 -12px 0 8px);
           border-radius: 50%;
+          // emulate flex="none"
+          -webkit-box-flex: 0;
+          -webkit-flex: 0 0 auto;
```
emoralesb05: You dont need to add the `webkit` prefixes since these are generated by the `autoprefixer` package

src/platform/core/chips/chips.component.html
```diff
@@ -11,7 +11,7 @@
                    (focus)="_handleChipFocus($event, chip)">
       <div class="td-chip" [class.td-chip-stacked]="stacked">
         <span class="td-chip-content">
-          <span *ngIf="!_chipTemplate?.templateRef">{{chip}}</span>
+          <span *ngIf="!_chipTemplate?.templateRef" class="text-truncate">{{chip}}</span>
```
emoralesb05: We tend to avoid using utility classes inside the components so we dont create a dependency to the external `SCSS` files, and the components can be used stand alone.

felipeik:
Hi @emoralesb05 , I just pushed the fixes to the comments!

felipeik:
> Awesome, thanks for the prompt response!
> 
> Do we still need to add the css rules to emulate `text-truncate` inside the component?

you mean this code?

```
     // truncate content on flexbox children
      min-width: 0;
```

this will be needed if we add text-truncate through a template

the text will not truncate correctly because of flexbox

but the developer will be able to add it manually if they wish, they will need to add this to the project CSS

this is a CSS hack that can be removed from the Covalent code and added to any project

let me know if I should remove it

PR #1250, opened 2018-10-10
feat(chips): added required attribute

pv-arunadevi:
## Description
Added required attribute to the chips component.

### What's included?

1. Added required attribute to <td-chips> component. When the value is true, at least one chip should be available in the chips list.
2. Place holder text will have '*' appended
3. We will not prevent user from deleting the last chip but Chip input will be displayed in warn color. When at least one chip is added, chip input will be reset to its primary color.
4. Default value is false.

#### Test Steps
- [x] `npm run serve`


src/app/components/components/chips/chips.component.html
```diff
@@ -400,6 +400,41 @@
     </mat-tab>
   </mat-tab-group>
 </mat-card>
+<!-- chip with required attribute -->
+<mat-card>
+  <mat-card-title>Chips with required attribute</mat-card-title>
+  <mat-card-subtitle>"required" attribute is set to true when we need atleast one chip to be available. When the last chip is deleted, input field will be displayed in warn color</mat-card-subtitle>
```
jeremysmartt: typo, change to "at least"

src/platform/core/chips/README.md
```diff
@@ -22,6 +22,10 @@ Leverage the templates to create your own chip or contact chip.
   + Set input position before or after the chips. Defaults to 'after'.
 + placeholder?: string
   + Placeholder for the autocomplete input.
++ required?: boolean
+  + Mandates atleast one chip to be available in the component. Appends * to the placeholder text.
```
jeremysmartt: "at least" here also

src/platform/core/chips/chips.component.ts
```diff
@@ -150,6 +151,19 @@ export class TdChipsComponent extends _TdChipsMixinBase implements IControlValue
     return this._requireMatch;
   }
 
+  /**
+   * required?: boolean
+   * Value is set to true if atleast one chip is needed
```
jeremysmartt: "at least" here also

src/app/components/components/chips/chips.component.html
```diff
@@ -400,6 +400,41 @@
...
+  <mat-divider></mat-divider>
+  <mat-tab-group mat-stretch-tabs dynamicHeight>
+    <mat-tab>
+      <ng-template matTabLabel>Demo</ng-template>
+      <div class="push">
+        <div class="mat-body-1">Type any text and press enter:</div>
+        <td-chips placeholder="Enter any string" required="true"></td-chips>
```
JulieMarie: This can be just `required`.

src/app/components/components/chips/chips.component.html
```diff
@@ -400,6 +400,41 @@
...
+      </div>
+    </mat-tab>
+    <mat-tab>
+      <ng-template matTabLabel>Code</ng-template>
+      <mat-card-content>
+        <p>HTML:</p>
+        <td-highlight lang="html">
+          <![CDATA[
+          <td-chips placeholder="Enter any string" required="true"></td-chips>
```
JulieMarie: Same here....just need `required`.

src/platform/core/chips/README.md
```diff
@@ -65,7 +69,7 @@ export class MyModule {}
 Example for HTML usage:
 
 ```html
-<td-chips placeholder="placeholder"
+<td-chips placeholder="placeholder" required="true"
```
JulieMarie: Same here....just `required`.  Similar to `requireMatch` and `stacked`.

pv-arunadevi:
All the review comments are implemented. @jeremysmartt @JulieMarie Please take a look again.

JulieMarie:
Awesome job Aruna!

src/platform/core/chips/chips.component.ts
```diff
@@ -420,6 +444,9 @@ export class TdChipsComponent extends _TdChipsMixinBase implements IControlValue
      * add a debounce ms delay when reopening the autocomplete to give it time
      * to rerender the next list and at the correct spot
      */
+    // Reset the color to primary
+    this.color = 'primary';
```
emoralesb05: So the problem here is that we are assuming that the reset is always primary.. that may not be the case..

I rather we piggyback on the `ng-invalid` class that's set when the component is invalid and change the colors of the underline when its set.

e.g. in _chips-theme.scss
https://github.com/Teradata/covalent/blob/develop/src/platform/core/chips/_chips-theme.scss#L70
```
td-chips {
   ...
     &.mat-warn,
    &.ng-invalid {
      .mat-form-field-underline .mat-form-field-ripple {
        background-color: mat-color($warn);
      }
    }
    &.ng-invalid {
      .mat-form-field-label-wrapper > label.mat-form-field-label.mat-form-field-empty {
        color: mat-color($warn);
      }
    }
```

src/platform/core/chips/chips.component.ts
```diff
@@ -456,6 +483,11 @@ export class TdChipsComponent extends _TdChipsMixinBase implements IControlValue
      */
     if (index === (this._totalChips - 1) && index === 0) {
       this._inputChild.focus();
+
+      // set the input field to warn color when the last chip is deleted when the required attribute is set
```
emoralesb05: This would be removed if we do what i mentioned on the previous comment.

PR #1255, opened 2018-10-15
fix(breadcrumbs): consistently align separator icons independent of their font-set class

### LINKED ISSUE #1254: bug(breadcrumbs): separator icon not correctly aligned, when associated to a font-set class different than 'material-icons'
#### Bug Report

When using a default font-set class different from 'material-icons', the separator icon wont be correctly aligned to the text.

Icon should be correctly assigned independent of the font-set class to which it belongs.

@angular/material,cdk@6.4.7
@covalent/core: 2.0.0-beta.3

##### Other information
Problem is caused by [this css selector in the source](https://github.com/Teradata/covalent/blob/v2.0.0-beta.3/src/platform/experimental/breadcrumbs/breadcrumb/breadcrumb.component.scss#L14), as it restricts the icon to have the 'material-icons' font-set class.

jotatoledo:
* Add new css class
* Reduces specificity of css selector used to style
the separator icon. Now the style is applied to the newly added class.
This makes the styling independent of the font-set class
associated to the icon, which is currently not the case.

Closes #1254 

#### Test Steps
- [ ] `npm run serve`
- [ ] then this
- [ ] finally this

### DIFF
--- a/src/platform/experimental/breadcrumbs/breadcrumb/breadcrumb.component.html
+++ b/src/platform/experimental/breadcrumbs/breadcrumb/breadcrumb.component.html
@@ -1,6 +1,7 @@
 <span *ngIf="displayCrumb" class="td-breadcrumb">
   <ng-content></ng-content>
   <mat-icon *ngIf="_displayIcon"
+            class="td-breadcrumb-icon"
             [style.cursor]="'default'"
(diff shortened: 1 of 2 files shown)


jeremysmartt:
good one @jotatoledo 👍

Issue #1232, opened 2018-08-30
Infinite scroll component missing in covalent lib


pv-arunadevi:
#### Feature Request

Infinite scroll would be a great value add in covalent lib. 

#### Bug Report

StackBlitz template: https://stackblitz.com/edit/covalent

#### Which version of Angular and Material, and which browser and OS does this issue affect?

Did this work in previous versions of Angular / Material?

emoralesb05:
Will be added in next release.

PR #1259, opened 2018-10-24
feat(virtual_scroll): added bottom event for infinite scrolling feature

felipeik:
## Description
Added Infinite Scroll functionality to Virtual Scroll

### What's included?
- Added `bottom` output to VirtualScroll so the parent knows when to load more items
- Added demo and documentation in the Virtual Scroll page
- Unit Tests

#### Test Steps
- [ ] npm run serve
- [ ] Open http://localhost:4200/#/components/virtual-scroll
- [ ] See docs rendering like usual
- [ ] Find Infinite Scroll card
- [ ] Scroll bottom and see it loading new items
- [ ] Click SCROLL TO FIRST
- [ ] Click SCROLL TO LAST
- [ ] See it loading new items again
- [ ] Click on the Code icon and see code example in the same card (HTML and JS)
- [ ] Added the `bottom` event documentation in the same page
- [ ] Added `bottom` to usage section at the bottom of the page

- [ ] npm run serve:prod still works.
- [ ] npm run tslint passes.
- [ ] npm run stylelint passes.
- [ ] npm test passes and code coverage is not lower.
- [ ] npm run build:lib still works.

Implementation for #1232


src/platform/core/virtual-scroll/virtual-scroll-container.component.ts
```diff
@@ -190,11 +205,32 @@ export class TdVirtualScrollContainerComponent implements AfterViewInit, AfterVi
...
     // that the changes will be marked
     Promise.resolve().then(() => {
       this._changeDetectorRef.markForCheck();
     });
   }
 
+  /**
+   * Method to emit bottom event if scrolled to the end of the list.
+   */
+  private bottomCheck(): void {
```
emoralesb05: Instead of checking on every single AfterViewChecked if the last object is the last rendered object.. we can just add  something to the `scroll` event to see if the bottom has been reached to be a bit more performant.

Also, its triggering the event when the last object is added to the dom.. not really when it hits the bottom.. so this way we know for sure its only going to trigger it then.

emoralesb05: e.g.

```
@HostListener('scroll', ['$event'])
  handleScroll(event: Event): void {
    let element: HTMLElement = (<HTMLElement>event.target);
    if (element) {
      let verticalScroll: number = element.scrollTop;
      if (this._scrollVerticalOffset !== verticalScroll) {
        this._scrollVerticalOffset = verticalScroll;
        if (this._initialized) {
          this._calculateVirtualRows();
        }
      }
      // check to see if bottom was hit to throw the bottom event
      if ((this._data.length * this.rowHeight) - (verticalScroll + this._hostHeight) === 0) {
        this.bottom.next();
      }
    }
  }
```

emoralesb05:
Illl merge it and create a PR to fix the comments since i wanna release this in our next beta.

Issue #1220, opened 2018-08-14
[Bug]: file-input doesnt handle disabled state via reactive forms.


emoralesb05:
#### Bug Report

File input needs to be able to be disabled when using reactive forms.

PR #1267, opened 2018-10-26
Handle file-input disabled state via reactive forms

n370:
## Description
Changes added here modify the `<td-file-input>` being used inside the `<td-dynamic-file-input>` component to use a `[disabled]` custom input property with the appropriate value to make it look and behave as the rest of the elements inside the dynamic component when it's in a *disabled* state.

### What's included?

- A fix for #1220

#### Test Steps

- [ ] Run `npm run serve`
- [ ] Navigate to http://localhost:4200/#/components/dynamic-forms
- [ ] Add a new file-input element to the "Generated Form" panel using the "Form Builder" panel.
- [ ] Toggle the "disable" switch to set the disable property on and off in the component and then hit the "Update Form" button to see the changes reflected on the dynamically generated component.

screen shot 2018-10-26 at 19 40 11

### DIFF
--- a/src/platform/dynamic-forms/dynamic-elements/dynamic-file-input/dynamic-file-input.component.html
+++ b/src/platform/dynamic-forms/dynamic-elements/dynamic-file-input/dynamic-file-input.component.html
@@ -27,6 +27,7 @@
   </button>
   <td-file-input class="td-file-input"
                   #fileInput
+                  [disabled]="control?.disabled"
                   [formControl]="control">
     <mat-icon>folder</mat-icon>
     <span>{{ label }}</span>


emoralesb05:
This wont fix this.. you cant use template driver forms combined with reactive forms anymore

image

hence why the https://github.com/Teradata/covalent/issues/1220 issue states to make it work with reactive forms.. which requires code changes in the components such as injecting `ngModel` optionally and such.

n370:
@felipeik I'm interested in keep working on this by my own as an independent open source contributor if you don't mind.

@emoralesb05 thanks for the feedback, I'll keep working on that. Expect me to be working on this over the next couple of days.

n370:
I've been very busy lately I don't think I'd be able to meet what I promised, sorry about that.

PR #1275, opened 2018-11-13
feat(breadcrumbs): remove not needed elements from breadcrumbs

emoralesb05:
## Description

Make sure we remove unnecessary elements from the `td-breadcrumbs` DOM.

e.g.

before
```
<td-breadcrumbs>
  <div class="td-breadcrumbs">
    <td-breadcrumb>
      <span class="td-breadcrumb">
```

after
```
<td-breadcrumbs class="td-breadcrumbs">
    <td-breadcrumb class="td-breadcrumb">
```

#### Test Steps

- [x] `npm run serve`
- [x] See demo still working properly

### DIFF
--- a/src/platform/core/breadcrumbs/breadcrumb/breadcrumb.component.ts
+++ b/src/platform/core/breadcrumbs/breadcrumb/breadcrumb.component.ts
@@ -1,7 +1,6 @@
 import {
   Component,
   ElementRef,
-  Renderer2,
   HostBinding,
   AfterViewInit,
   ChangeDetectionStrategy,
@@ -12,6 +11,10 @@ import {
   selector: 'td-breadcrumb, a[td-breadcrumb]',
   styleUrls: ['./breadcrumb.component.scss'],
   templateUrl: './breadcrumb.component.html',
+  /* tslint:disable-next-line */
+  host: {
+    class: 'mat-button td-breadcrumb',
+  },
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class TdBreadcrumbComponent implements AfterViewInit {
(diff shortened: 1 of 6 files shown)


src/platform/core/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -47,7 +48,11 @@ export class TdBreadcrumbsComponent implements OnInit, DoCheck, AfterContentInit
    */
   @Input('separatorIcon') separatorIcon: string = 'navigate_next';
 
-  constructor(private _elementRef: ElementRef, private _changeDetectorRef: ChangeDetectorRef) { }
+  constructor(private _elementRef: ElementRef,
+              private _renderer: Renderer2,
```
jotatoledo: Same comment as before

src/platform/core/breadcrumbs/breadcrumb/breadcrumb.component.ts
```diff
@@ -56,6 +56,7 @@ export class TdBreadcrumbComponent implements AfterViewInit {
               private _renderer: Renderer2,
               private _changeDetectorRef: ChangeDetectorRef) {
     this._renderer.addClass(this._elementRef.nativeElement, 'mat-button');
+    this._renderer.addClass(this._elementRef.nativeElement, 'td-breadcrumb');
```
jotatoledo: The dependency on `Renderer2` could be removed by binding the static classes through the `host` attribute of the Component decorator. IMO is a lot cleaner, as `Renderer2` is a really low level API, which is kinda overkill for this simple use case.

emoralesb05: Yeah, that makes sense. I typically avoid using `host` because of change detection but in this case is static so its just easier to use it.

PR #1277, opened 2018-11-13
fix(dynamic-forms):  added missing hints for checkboxes, slide-toggles and sliders

JulieMarie:
## Description
Added hints in for those elements that don't have built in `<mat-hint></mat-hint>` (ie. checkboxes, slide-toggles and sliders).

### What's included?
- Added hints to checkboxes
- Added hints to slide-toggles (booleans)
- Added hints to sliders

#### Test Steps
- [ ] `npm run serve`
- [ ] Go to http://localhost:4200/#/components/dynamic-forms and verify that hints show under `Dynamic Number Elements` and `Dynamic Boolean Elements`

screen shot 2018-11-13 at 1 06 19 pm

### DIFF
--- a/src/platform/dynamic-forms/dynamic-forms.component.scss
+++ b/src/platform/dynamic-forms/dynamic-forms.component.scss
@@ -3,6 +3,11 @@
     .mat-form-field-infix {
       width: auto;
     }
+    .hint {
+      color: rgba(0, 0, 0, 0.54);
+      font-size: 75%;
+      display: block;
+    }
   }
   // [layout-wrap]
   flex-wrap: wrap;
@@ -21,11 +26,5 @@
     box-sizing: border-box;
     position: relative;
     padding: 4px 4px 8px;
-
-    .hint {
-      color: rgba(0, 0, 0, 0.54);
-      font-size: 10.5px;
-      display: block;
-    }
   }
 }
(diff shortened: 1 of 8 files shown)


src/platform/dynamic-forms/dynamic-forms.component.scss
```diff
@@ -3,6 +3,11 @@
     .mat-form-field-infix {
       width: auto;
     }
+    .hint {
```
emoralesb05: We should create a more specific name than just`hint`. Something like `td-dynamic-element-hint` or something

src/platform/dynamic-forms/dynamic-forms.component.scss
```diff
@@ -3,6 +3,11 @@
     .mat-form-field-infix {
       width: auto;
     }
+    .td-dynamic-elemenet-hint {
+      color: rgba(0, 0, 0, 0.54);
```
emoralesb05: Just noticed this.. we might need to get something from the theme to change this since in dark mode it needs to change its contrast.

we can probably piggy back on the class material uses for their hints

image

JulieMarie: Good catch....fixed now.

PR #1278, opened 2018-11-14
perf(breadcrumbs): simplify logic to show/hide breadcrumbs 

### LINKED ISSUE #1274: bug(breadcrumbs): potential infinite loops
#### Bug Report

The current implementation of `TdBreadcrumbs` is prone to fall into an infinite loop under certain conditions.

For example in [this blitz](https://stackblitz.com/edit/covalent-breadcrumb-infiniteloop):

If the array bounded to the `ngFor` (line 27)  is the `breadcrumbs` member in the component, it works fine. But if the bounded array is `getterBreadcrumbs`, then the app will fall into an infinite loop (view will freeze), as the bounded array is recalculated during change detection which triggers a change detection at the `TdBreadcrumbs` component level.

Although bounding getters that returns new references to the template isnt a good pattern, this shouldnt result in the app transitioning into an infinite loop state.

angular: 7.x
materia/cdk: 7.x
covalent/core: 2.0.0-beta.4

#### Did this work in previous versions of Angular / Material?

The bug was not present on covalent/core: 2.0.0-beta.3 with angular,cdk,material: 6.x

emoralesb05:
also change icon to `chevron_right` and change usage of Promise to setInterval to fix issue in https://github.com/Teradata/covalent/issues/1274

### What's included?

- Simplify logic in breadcrumbs hide/show logic to a single for loop
- fixed infinity loop issue

closes #1274


src/platform/core/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -128,41 +128,26 @@ export class TdBreadcrumbsComponent implements OnInit, DoCheck, AfterContentInit
...
-          break;
-        }
+    let crumbWidthSum: number = 0;
+    let hiddenCrumbs: TdBreadcrumbComponent[] = [];
+    // loop through crumbs in reverse order to calculate which ones should be removed
+    for (let i: number = crumbsArray.length - 1; i >= 0; i--) {
+      let breadcrumb: TdBreadcrumbComponent = crumbsArray[i];
+      // if crumb exceeds width, then we skip it from the sum and add it into the hiddencrumbs array
+      // and hide it
+      if ((crumbWidthSum + breadcrumb.width) > this.nativeElementWidth) {
```
jotatoledo: Assuming that `this.nativeElementWidth` remains constant once this operation starts, this could be further simplified into something like:

```
...
const availableWidth = this.nativeElementWidth;
...
for(....){
 let breadcrumb = ....;
 let breadcrumbWillOverflow = availableWidth - breadcrumb.width < 0;
 if(breadcrumbWillOverflow){
  ...
 }
 else
 {
   ...
 }
 availableWidth -= breadcrumb.width;
}
```

PR #1303, opened 2018-11-27
fix(data-table): Cell alignment 

### LINKED ISSUE #871: Data Table cell alignment

Bug/Regression (?)

#### Feature Request

Require cell alignment in table. Was working.

#### Bug Report

Previously developer was able to specify the text-alignment for a cell. With the new `td-data-table-cell-content-wrapper` div this is no longer possible. 

Developers can add alignment to the table cells.

Need ability to modify layout of the contents contained within a data table.

Material 2.0.0-beta.10
Angular 4.3.6
Covalent 1.0.0-beta.7

Yes

philgetz:
### Description
Added properties 'cellAlign' on atomic components TdDataTableColumn and TdDataTableCell to align cell content.

Fixes #871 

### What's included?

- align: input property

#### Test Steps
- [ ] Open file covalent/src/app/components/components/data-table/data-table.component.html
- [ ]   (Line 30) Change property align="start" to align="end"
- [ ] `npm run serve`
- [ ] Open http://localhost:4200/#/components/data-table
- [ ] Scroll to card  **Custom Data Table** on "comments" column, check cells content aligns to 'end'
- [ ] In toolbar click on **More - Icon button** (Last button from left to right, next to notification button)
- [ ] Click on LTR
- [ ] On "comments" column from Step 5, check cell content still aligns to 'end' in the context of RTL

### DIFF
--- a/src/platform/core/data-table/data-table-cell/data-table-cell.component.ts
+++ b/src/platform/core/data-table/data-table-cell/data-table-cell.component.ts
@@ -2,8 +2,8 @@ import { Component, Input, Renderer2, ElementRef, HostBinding } from '@angular/c
 
 export enum TdDataTableCellAlign {
   Center = 'center',
-  Left = 'left',
-  Right = 'right',
+  Start = 'start',
+  End = 'end',
 }
 
 @Component({
@@ -24,7 +24,7 @@ export class TdDataTableCellComponent {
   @Input('numeric') numeric: boolean = false;
 
   /**
-   * cellAlign?: 'center' | 'right' | 'left'
+   * cellAlign?: 'start' | 'center' | 'end'
    * Makes cell content align on demand
    * Defaults to 'left', overrides numeric
    */
--- a/src/platform/core/data-table/data-table-cell/data-table-cell.component.ts
+++ b/src/platform/core/data-table/data-table-cell/data-table-cell.component.ts
@@ -1,11 +1,5 @@
 import { Component, Input, Renderer2, ElementRef, HostBinding } from '@angular/core';
 
-export enum TdDataTableCellAlign {
-  Center = 'center',
-  Start = 'start',
-  End = 'end',
-}
-
 @Component({
   /* tslint:disable-next-line */
   selector: 'td[td-data-table-cell]',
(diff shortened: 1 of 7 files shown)


kyleledbetter:
are we considering RTL for the alignment?

philgetz:
> are we considering RTL for the alignment?

Just committed to work with RTL.

src/platform/core/data-table/data-table-cell/data-table-cell.component.ts
```diff
@@ -1,5 +1,11 @@
 import { Component, Input, Renderer2, ElementRef, HostBinding } from '@angular/core';
 
+export enum TdDataTableCellAlign {
```
emoralesb05: You can probably just change this to a `type` variable to make it easier to use. We are trying to go away from enums

e.g.
```
export type TdDataTableCellAlign = 'start' | 'center' | 'end';
```

src/platform/core/data-table/data-table-cell/data-table-cell.component.ts
```diff
@@ -8,13 +14,28 @@ import { Component, Input, Renderer2, ElementRef, HostBinding } from '@angular/c
...
    */
   @Input('numeric') numeric: boolean = false;
 
+  /**
+   * cellAlign?: 'start' | 'center' | 'end'
+   * Makes cell content align on demand
+   * Defaults to 'left', overrides numeric
+   */
+  @Input() 
+  set cellAlign(cellAlign: TdDataTableCellAlign) {
```
emoralesb05: From a usage standpoint, since its directly applied to the cell, having it called `cellAlign` is verbose.

It will be simpler to just call it `align`.. it also provides the same API that a normal HTML table has.

philgetz: Renamed to align on latest commit

src/platform/core/data-table/data-table-cell/data-table-cell.component.ts
philgetz: Switching to type from enum in latest commit 98cadfe

PR #1350, opened 2019-02-21
fix(build): Drop support for yarn in favor of package-lock.json

richavyas:
## Description
`ng serve` after  **yarn install** started failing because parse5 dependency for @angular/cdk/schematics isn't being resolved correctly when installed using yarn but works fine with npm install, it throws the following compile error with yarn:

```
ERROR in node_modules/@angular/cdk/schematics/ng-update/html-parsing/elements.d.ts(8,10): error TS2305: Module '"/Users/rv186024/projects/covalent/node_modules/parse5/lib/index"' has no exported member 'DefaultTreeElement'.
node_modules/@angular/cdk/schematics/utils/html-head-element.d.ts(9,10): error TS2305: Module '"/Users/rv186024/projects/covalent/node_modules/parse5/lib/index"' has no exported member 'DefaultTreeElement'.
node_modules/@angular/cdk/schematics/utils/parse5-element.d.ts(8,10): error TS2305: Module '"/Users/rv186024/projects/covalent/node_modules/parse5/lib/index"' has no exported member 'DefaultTreeElement'.
```
As part of this PR, package-lock.json is added for npm versions and yarn.lock is removed. 

Note: Another way to fix this is by avoiding type checking of declaration files (files with extension .d.ts) of libraries. This can be disabled by specifying `skiplibcheck : true` compiler option in tsconfig.js as mentioned [here](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#new---skiplibcheck). This will also make our build faster. The dependencies added as part of this PR should be removed if we choose to go with `skiplibcheck : true` option in future.

### What's included?
- Added package-lock.json and removed yarn.lock file

#### Test Steps
- [x] rm -rf node_modules
- [x] npm install
- [x] ng serve and see everything is working
- [x] rm -rf node_modules
- [x] yarn install
- [x] ng serve and see everything is working


jeremysmartt:
Can you rename this PR also, "Drop support for yarn.lock in favor of package-lock.json"  Or something like that?

PR #1355, opened 2019-02-26
feat(highlight) deprecate lang input in favor of codeLang

stevenov7:
## Description

This PR merges into a feature branch: `feature/a11y-code-fixes`

Batch-1: Of code related a11y issue fixes reported by pa11y, Jenn is resolving the visual side of a11y/pa11y issues in a different feature branch.

### What's included?

- [ ] td-highlight language input alias changed from lang to codeLang, depreciate lang input alias

#### Test Steps

- [ ] Primarily want code review b/c will take a long time to run `npm run a11y` every batch PR. When all the batches get merged into feature branch - the last batch PR we will run `npm run a11y` and it should not report any issues.
- [ ] Sanity check
  - [ ] `npm run serve`

### DIFF
--- a/src/platform/highlight/highlight.component.ts
+++ b/src/platform/highlight/highlight.component.ts
@@ -33,14 +33,14 @@ export class TdHighlightComponent implements AfterViewInit {
   }
 
   /**
-   * lang?: string
+   * codeLang?: string
    *
    * Language of the code content to be parsed as highlighted html.
    * Defaults to `typescript`
    *
    * e.g. `typescript`, `html` , etc.
    */
-  @Input('lang') language: string = 'typescript';
+  @Input('codeLang') language: string = 'typescript';
 
   /**
    * contentReady?: function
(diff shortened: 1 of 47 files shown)


src/platform/highlight/highlight.component.ts
```diff
@@ -33,14 +33,20 @@ export class TdHighlightComponent implements AfterViewInit {
   }
 
   /**
-   * lang?: string
+   * Deprecated: due to a11y issues reported by pa11y
+   * Please see @Input('codeLang') as its replacement
+   */
+  @Input('lang') languageDeprecated: string = 'typescript';
```
christianmemije: If you pass in the deprecated `lang` prop, it ignores the prop and uses `'typescript` instead.

stevenov7:
@christianmemije what do you suggest the intended behavior be? Do you think an error should throw if  someone tries to use both properties at the same time?

PR #1368, opened 2019-03-19
feat(): added fullscreen directive to be place on element 

joejcarey:
## Description
Added a fullscreen directive to be place on an element. 

### What's included?
- new directive directory
- fullscreen directive to be placed on a button  

#### Test Steps
- [ ] assign a template reference variables to the element you want full screen
- [ ] use `tdFullScreen` on the fullscreen action button and input that template reference
- [ ] press button and confirm that targeted div is fullscreen
- [ ] add the `tdEscapeKey="Tab"` or other keydown event key string and confirm that it exits the fullscreen as well. 

**does not work on elements that remove the template reference such as `mat-card`.


src/platform/core/common/common.module.ts
```diff
@@ -11,13 +11,10 @@ import { FormsModule } from '@angular/forms';
 // Form Directives
 import { TdAutoTrimDirective } from './forms/auto-trim/auto-trim.directive';
 
-const TD_FORMS: Type<any>[] = [
```
christianmemije: I wonder if TD_FORMS should be renamed to TD_ DIRECTIVES, and include the TdFullscreenDirective inside this for consistency.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,86 @@
+import { Directive, HostListener, Input, ElementRef } from '@angular/core';
+
+@Directive({
+  selector: '[tdFullScreen]',
+})
+export class TdFullscreenDirective {
+  private _webview: any;
+
+  @Input() tdFullScreen: HTMLDivElement;
+  @Input() tdEscapeKey?: String;
```
christianmemije: I don't really see a use case for passing in a custom escape key since `esc` is the web standard.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  /**
+   * showFullScreenEditor request for full screen of Code Editor based on its browser type.
+   */
+
+  private showFullScreenEditor(): void {
+    if (this.tdFullScreen) {
+      if (this._webview) {
```
christianmemije: No webview logic should be in here.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,86 @@
...
+  @HostListener('document:keydown', ['$event']) onKeydownHandler(
+    event: KeyboardEvent,
+  ): void {
+    if (event.key === this.tdEscapeKey) {
+      this.exitFullScreenEditor();
+    }
+  }
+
+  /**
+   * showFullScreenEditor request for full screen of Code Editor based on its browser type.
```
christianmemije: 🙅‍♂️

christianmemije:
I think a cleaner api would be

```html
<div 
  tdFullScreenRef]="ref"  
  [(tdInFullScreen)]="inFullScreen"
>
</div>
```

This makes it so you don't rely on an element click event. You are also aware of what the fullscreen state is by making `tdInFullScreen` have 2-way binding

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,86 @@
+import { Directive, HostListener, Input, ElementRef } from '@angular/core';
+
+@Directive({
+  selector: '[tdFullScreen]',
+})
+export class TdFullscreenDirective {
+  private _webview: any;
+
+  @Input() tdFullScreen: HTMLDivElement;
```
emoralesb05: We can just inject `ElementRef` instead of passing the element. Assuming the directive gets set on the element you wanna add fullscreen capabilities.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,86 @@
+import { Directive, HostListener, Input, ElementRef } from '@angular/core';
+
+@Directive({
+  selector: '[tdFullScreen]',
```
emoralesb05: We should add an `exportAs` so people can reference the methods programmatically.

Then the button can just do something like `myElement.fullscreen()`

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,66 @@
+import { Directive, HostListener, Input, ElementRef, Inject } from '@angular/core';
+import { DOCUMENT } from '@angular/common';
+@Directive({
+  selector: '[tdFullScreen]',
+  exportAs: 'tdFullScreen',
+})
+export class TdFullscreenDirective {
+  @Input() tdEscapeKey?: String;
```
christianmemije: I think we should remove this input.

src/platform/core/common/directives/fullscreen/fullscreen.directive.spec.ts
```diff
@@ -0,0 +1,93 @@
+import { TestBed, ComponentFixture } from '@angular/core/testing';
+import { Component, DebugElement } from '@angular/core';
+import { TdFullscreenDirective } from './fullscreen.directive';
+import {
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting
+} from '@angular/platform-browser-dynamic/testing';
+import { By } from '@angular/platform-browser';
+
+@Component({
```
christianmemije: Can we get some tests where we have multple elements will fullscreen directives and test the fullscreen toggling interaction.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,66 @@
...
+  @HostListener('document:keydown', ['$event']) public onKeydownHandler(
+    event: KeyboardEvent,
+  ): void {
+    if (event.key === this.tdEscapeKey) {
+      this.exitFullScreen();
+    }
+  }
+
+  @HostListener('document:fullscreenchange') fsChangeHandler(): void {
+    this.fullScreenIsActive = !this.fullScreenIsActive;
```
christianmemije: If we have multiple directives on a page then all of them will have the same value. So after we listen to this event, we need to check if there is a currently full screen element. If so, we need to check whether the element that is fullscreen is this element. If so, then we set fullScreenIsActive to true.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,66 @@
...
+    for (const handler of Object.keys(enterFullScreenMap)) {
+      if (element[handler]) {
+        enterFullScreenMap[handler]();
+      }
+    }
+  }
+
+  public exitFullScreen(): void {
+    const exitFullScreenMap: object = {
+      // Chrome
```
christianmemije: Lets make this map have the same order as the previous one.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
emoralesb05: Nice catch, i think its just easier to set the flag when calling the methods.

src/platform/core/common/directives/fullscreen/fullscreen.directive.spec.ts
kriswinbush: good call

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
kriswinbush: agreed, will move the state change into the method calls

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    for (const handler of Object.keys(enterFullScreenMap)) {
+      if (element[handler]) {
+        enterFullScreenMap[handler]();
+      }
+    }
+  }
+
+  public exitFullScreen(): void {
+    const exitFullScreenMap: object = {
```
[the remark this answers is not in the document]
kriswinbush: conflicts with tslint:recommended 'ban-types' rul, https://github.com/palantir/tslint/blob/master/src/configs/recommended.ts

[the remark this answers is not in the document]
emoralesb05: ah kk, the one in the enterFullscreen is `Object`, it needs to change to `object`.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,66 @@
...
+    this.fullScreenIsActive = !this.fullScreenIsActive;
+  }
+
+  public toggleFullScreen(): void {
+    this.fullScreenIsActive ? this.exitFullScreen() : this.enterFullScreen();
+  }
+
+  public enterFullScreen(): void {
+    const element: HTMLElement = this.el.nativeElement;
+    const enterFullScreenMap: Object = {
```
emoralesb05: We should change this to `object`.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -1,64 +1,56 @@
...
-    if (event.key === this.tdEscapeKey) {
-      this.exitFullScreen();
-    }
-  }
+  constructor(@Inject(DOCUMENT) private _document: FsDocument, private _el: ElementRef) {}
 
-  @HostListener('document:fullscreenchange') fsChangeHandler(): void {
-    this.fullScreenIsActive = !this.fullScreenIsActive;
+  @HostListener('document:fullscreenchange', ['$event']) public fsChangeHandler(event: Event): void {
+    this.fullScreenIsActive = event.srcElement === this._document.fullscreenElement ? true : false;
```
christianmemije: Can we get rid of the event and just do `this.fullScreenIsActive = this._document.fullscreenElement === this._el.nativeElement`?

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -1,64 +1,56 @@
...
-      // Chrome
-      requestFullscreen: () => element.requestFullscreen(),
-      // Safari
-      webkitRequestFullscreen: () => (<any>element).webkitRequestFullscreen(),
-      // IE
-      msRequestFullscreen: () => (<any>element).msRequestFullscreen(),
-      // Firefox
-      mozRequestFullScreen: () => (<any>element).mozRequestFullScreen(),
+    const { _document: { fullscreenElement }, _el: { nativeElement } }: TdFullscreenDirective = this;
+    const enterFullScreenMap: object = {
```
christianmemije: Use the `FsDocument` interface here?

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,58 @@
...
+      requestFullscreen: () => nativeElement.requestFullscreen(), // Chrome
+      webkitRequestFullscreen: () => nativeElement.webkitRequestFullscreen(), // Safari 
+      mozRequestFullScreen: () => nativeElement.mozRequestFullScreen(), // Firefox
+      msRequestFullscreen: () => nativeElement.msRequestFullscreen(), // IE
+    };
+
+    for (const handler of Object.keys(enterFullScreenMap)) {
+      if (nativeElement[handler] && !fullscreenElement) {
```
christianmemije: We want to enter fullscreen even if there is already another element that is fullscreen.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
kriswinbush: without the event ng wont know when the document.fullscreenElement is updated.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
kriswinbush: Is there a use case were we want to nest a fullscreen element inside of and activated fullscreen element?

src/platform/core/common/directives/fullscreen/fullscreen.directive.spec.ts
```diff
@@ -9,14 +9,16 @@ import { By } from '@angular/platform-browser';
 
 @Component({
   template: `<div tdFullScreen #myDirective="tdFullScreen" tdEscapeKey="Tab">
```
emoralesb05: We should probably remove the `tdEscapeKey` from here also.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,80 @@
...
+    };
+
+    for (const handler of Object.keys(exitFullScreenMap)) {
+      if (_document[handler] && this.tdFullScreenElement() === nativeElement) {
+        exitFullScreenMap[handler]();
+      }
+    }
+  }
+
+  public tdFullScreenElement() {
```
christianmemije: Could we rename this to `getFullScreenElement()`?

src/app/components/components/directives/directives.component.html
```diff
@@ -50,3 +50,38 @@ <h3>Setup:</h3>
     </td-highlight>
   </mat-card-content>
 </mat-card>
+<mat-card tdFullScreen #tdFullScreen="tdFullScreen">
+  <mat-card-header layout="row">
+    <mat-card-title flex>Fullscreen directive</mat-card-title>
+    <div flex></div>
+    <button *ngIf="tdFullScreen.fullScreenIsActive" mat-button color="primary" (click)="tdFullScreen.exitFullScreen()">exit fullscreen</button>
```
christianmemije: Let's only have one exit fullscreen button. And maybe put both the enter and exit in the mat-card-actions.

src/platform/core/common/directives/fullscreen/fullscreen.directive.ts
```diff
@@ -0,0 +1,80 @@
...
+    };
+
+    for (const handler of Object.keys(exitFullScreenMap)) {
+      if (_document[handler] && this.getFullScreenElement() === nativeElement) {
+        exitFullScreenMap[handler]();
+      }
+    }
+  }
+
+  private getFullScreenElement() {
```
emoralesb05: You need to add `_` at the beginning of private members

src/app/components/components/directives/directives.component.html
```diff
@@ -50,3 +50,37 @@ <h3>Setup:</h3>
     </td-highlight>
   </mat-card-content>
 </mat-card>
+<mat-card tdFullScreen #tdFullScreen="tdFullScreen">
+  <mat-card-header layout="row">
+    <mat-card-title flex>Fullscreen directive</mat-card-title>
+  </mat-card-header>
+  <mat-divider></mat-divider>
+  <mat-card-content style="background-color: white;">
```
emoralesb05: Any reason for this background color?

PR #1396, opened 2019-06-08
feat(search): tweak styles/layout on appearance

emoralesb05:
## Description

Since now we expose appearance from `mat-form-field`.. we need to move the cancel button and tweak some styles depending on the appearance chosen to display a proper search layout.

#### Test Steps

- [ ] `npm run serve`
- [ ] Go to search demo
- [ ] See how styles change depending on appearance

### DIFF
--- a/src/platform/core/search/search-input/search-input.component.html
+++ b/src/platform/core/search/search-input/search-input.component.html
@@ -11,7 +11,18 @@
             (blur)="handleBlur()"
             (search)="stopPropagation($event)"
             (keyup.enter)="handleSearch($event)"/>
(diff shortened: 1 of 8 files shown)


src/platform/core/search/search-input/search-input.component.html
```diff
@@ -11,12 +11,23 @@
             (blur)="handleBlur()"
             (search)="stopPropagation($event)"
             (keyup.enter)="handleSearch($event)"/>
+    <span matSuffix *ngIf="appearance === 'fill' || appearance === 'outline' || appearance === 'standard'">
```
christianmemije: `['fill', 'outline', 'standard'].includes(appearance)` could make this more concise.

PR #1406, opened 2019-06-26
refactor(media): simplify service provider

jotatoledo:
* Since v6 angular provides a new way of registering singleton services
by means of the provide in root module pattern.


jotatoledo:
IMO the folder structure could also be flattened a bit. Having 2 folders with 1 file each is counterproductive.

emoralesb05:
finally merged! haha

PR #1407, opened 2019-06-26
refactor(data-table): simplify service provider

jotatoledo:
* Simplify how the singletone service is provided. Since v6 angular provides a new way of registering singleton services
by means of the provide in root module pattern.


emoralesb05:
Just FYI @jotatoledo, we will do a 2.1 release with fixes and some extra components and then immediately after that a 3.0.0 beta.0 with Angular 8 upgrade and the refactor PR's. Just waiting for that to get your refactor PR's in 😛

PR #1408, opened 2019-06-28
fix(tab-select): Mark for check when number of tabs change

andrelaszlo:
I discussed this super briefly with @emoralesb05 over email today.

When a `td-tab-option` is added to a `td-tab-select` component, the new tabs are sometimes not rendered.

I don't think the error is on my side - I have reproduced it in a simple [repo on Stackblitz][stackblitz], that shows a very simple example of adding tabs programmatically can trigger this behavior. This example works well with `mat-tab-group` as well.

Unfortunately, I haven't found a good way to write a failing test for this bug since change detection works differently while testing. If you have an idea for a good test, please let me know and I'll add it!

My current understanding is that since the component is using `OnPush`, change detection doesn't work when changes happen inside the `ContentChildren` observable. Either way, the fix is straightforward and seems to work. Unless you add/remove tabs like crazy, I don't think that it should trigger many extra change detections so the performance impact should be minimal.

[stackblitz]: https://stackblitz.com/edit/covalent-tab-select-component-bug-poc
[mat-tabs-group-spec]: https://github.com/angular/components/blob/27a08cc6017852c93d52a4f83ec7b7e587db928a/src/material/tabs/tab-group.spec.ts

### DIFF
--- a/src/platform/core/tab-select/tab-select.component.ts
+++ b/src/platform/core/tab-select/tab-select.component.ts
@@ -154,6 +154,7 @@ export class TdTabSelectComponent extends _TdTabSelectMixinBase
     this._values = this.tabOptions.map((tabOption: TdTabOptionComponent) => {
       return tabOption.value;
     });
+    this._changeDetectorRef.markForCheck();
   }
 
   /**


emoralesb05:
Awesome, thanks for this!

PR #1460, opened 2019-07-26
ci(commitlint): add commitlint

christianmemije:
### What's included?

- Adds a pre commit hook to validate commit messages using commitlint 
- Adds a travis script to verify commit messages. See the error that is thrown here https://travis-ci.org/Teradata/covalent/jobs/565596117 Not the most helpful message, but at least it says a commit message is malformed.
- Using [commitlint/config-angular](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-angular). Should we use that or [commitlint/config-conventional](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional)?

#### Test Steps

- [ ] `npm i`
- [ ] try to commit a message that is not of right format

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ before_script:
   - sh -e /etc/init.d/xvfb start
 jobs:
   include:
+    - env: 'MODE=commitlint'
+      script: commitlint-travis
     - env: 'MODE=prettier'
     - env: 'MODE=lint'
     - env: 'MODE=aot'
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -16,6 +16,8 @@ before_script:
   - sh -e /etc/init.d/xvfb start
 jobs:
   include:
+    - env: 'MODE=commitlint'
+      script: commitlint-travis
```
christianmemije: This works, but is this where we want to add this script?

PR #1461, opened 2019-07-26
chore(): upgrade to angular@8

emoralesb05:
## Description

Upgrade dependencies to latest angular and changed code base to fix breaking changes.
Also fixed npm audit.

#### Test Steps

- [ ] `npm install`
- [ ] `npm run serve`
- [ ] Make sure all components still work properly

NOTE: had to ignore some linting issues in the new lazy loading because of this
https://github.com/angular/angular-cli/issues/14690


src/app/app.routes.ts
```diff
@@ -17,37 +17,49 @@ const routes: Routes = [
     // preload: true loads the module immediately
     path: '',
     data: { preload: false },
-    loadChildren: './components/docs/docs.module#DocsModule',
+    loadChildren: () =>
+      /* tslint:disable-next-line */
```
christianmemije: Can we specify the exact rule we are disabling? Or use `any`, or do 
```ts
 import('./components/docs/docs.module').then((m: typeof import('./components/docs/docs.module')) => m.DocsModule),
```
😅

christianmemije: Oops, just read your comment on why `any` does not work.

kriswinbush:
could add enableIvy flag set to false in tsconfig.app.json, seem like we are pretty close to being able to compile with Ivy.
"angularCompilerOptions": {
    "enableIvy": false
  },

Issue #1438, opened 2019-07-15
Utility Functions - Copy to clipboard


jeremysmartt:

jeremysmartt:
@philgetz @christianmemije To Review

PR #1475, opened 2019-09-24
Utility Functions - Copy to clipboard

thomaslennan:
## Description

Helper function to copy text to system clipboard.

Fixes https://github.com/Teradata/covalent/issues/1438

### What's included?
- Reverted scaffolding for separate utility package
- Added 'functions' subdirectory under core/common
- Added and exported copyToClipboard function
- Added sample component that initially is used to document how to use copyToClipboard
function but subsequently can be utilized for other future functions.

#### Test Steps
- [ ] `npm run serve`
- [ ] Confirmed sample component content looks correct and copies to clipboard as expected
- [ ] Confirmed built core package included copyToClipboard function
- [ ] Confirmed function could be imported via '@covalent/core/common'

Issue1438


src/platform/core/common/functions/clipboard.ts
```diff
@@ -0,0 +1,25 @@
...
+  // Create a temporary textarea element and append to DOM
+  const fakeTextArea: HTMLTextAreaElement = document.createElement('textarea');
+  document.body.appendChild(fakeTextArea);
+
+  // Assign value to be copied to clipboard
+  fakeTextArea.value = value;
+  fakeTextArea.select();
+
+  // Copy to clipboard
+  document.execCommand('copy');
```
christianmemije: Can we return the value of this which is a boolean whether the action was successful?

thomaslennan: Done

src/app/components/components/functions/functions.component.html
```diff
@@ -0,0 +1,84 @@
...
+                <textarea matInput
+                  [(ngModel)]="copytext"
+                  placeholder="Text to be copied"
+                  rows="1"
+                  name="copytextinput">
+                </textarea>
+              </mat-form-field>
+            </div>
+          </form>
+          <button mat-button
```
christianmemije: Small suggestion would be to move this within the `mat-card-actions` section.

thomaslennan: Done

PR #1481, opened 2019-09-30
Update/refactor services in root 1401

kriswinbush:
## Description

update services to be providedIn: 'root' and/or providedIn: Module were applicable
### What's included?

updates to the following files
- CovalentCommonModule, icon.service.ts, router-path.service.ts
  (providedIn: CovalentCommonModule)
- CovalentDialogService, dialog.service.ts (providedIn: CovalentDialogModule)
- CovalentFileService, file.service.ts (providedIn: CovalentFileModule)
- CovalentSearchModule (removed unused ModuleWithProviders import)
- CovalentStepsModule (removed unused ModuleWithProviders import)
- CovalentDynamicFormsModule, dynamic-forms.service.ts (providedIn: 'root')
- CovalentBaseEchartsModule, chart-options.service.ts (providedIn: ‘root’)
- CovalentHighlightModule (removed ModuleWithProviders import)
- CovalentMarkdownModule (removed ModuleWithProviders import)
Non Modified Services
- Loading Module (factory provider)
- CovalentHttpModule (forRoot implementation)

#### Test Steps

- [ ] `npm run serve`
- [ ] open `localhost:4200` to the demo/documentation
- [ ] check that demo implementations of components with services are working correctly

- [ ] `npm run test` passes and code coverage is not lower.


src/platform/core/common/services/icon.service.ts
```diff
@@ -4,8 +4,10 @@
  */
 
 import { Injectable } from '@angular/core';
-
-@Injectable()
+import { CovalentCommonModule } from '../common.module';
```
stevenov7: Why is this one not in provided in root?

src/platform/dynamic-forms/services/dynamic-forms.service.ts
```diff
@@ -54,7 +54,9 @@ export interface ITdDynamicElementConfig {
 
 export const DYNAMIC_ELEMENT_NAME_REGEX: RegExp = /^[^0-9][^\@]*$/;
 
-@Injectable()
+@Injectable({
+  providedIn: 'root',
```
stevenov7: Hmm why is this one provided in root? and not the dynamic-forms module?

src/platform/echarts/base/chart-options.service.ts
```diff
@@ -2,7 +2,9 @@ import { Injectable, Provider, Optional, SkipSelf } from '@angular/core';
 
 import { Subject, Observable, BehaviorSubject } from 'rxjs';
 
-@Injectable()
+@Injectable({
+  providedIn: 'root',
```
stevenov7: Hmm why is this one provided in root? and not the dynamic-forms module?

emoralesb05: This actually has to be provided per component.. since its meant to manage the component state. So i think we should revert to what we had.

src/platform/dynamic-forms/services/dynamic-forms.service.ts
emoralesb05: This one looks like it doesnt have any dependencies, so its fine to be provided in `root`

src/platform/core/common/services/router-path.service.ts
```diff
@@ -2,8 +2,10 @@ import { Injectable } from '@angular/core';
 import { Router, RoutesRecognized } from '@angular/router';
 
 import { filter, pairwise } from 'rxjs/operators';
-
-@Injectable()
+import { CovalentCommonModule } from '../common.module';
+@Injectable({
+  providedIn: CovalentCommonModule,
```
emoralesb05: This can probably be provided at `root` since we probably just want a single instance.

src/platform/core/common/services/icon.service.ts
emoralesb05: Agree, this should be provided at `root`

kriswinbush: done

src/platform/dynamic-forms/services/dynamic-forms.service.ts
kriswinbush: no change

src/platform/echarts/base/chart-options.service.ts
kriswinbush: reverted

src/platform/core/common/services/router-path.service.ts
kriswinbush: done

PR #1505, opened 2019-10-04
Feat/md nav hide home

christianmemije:
## Description

- Hide home button on 1st and 2nd levels.
- Show title of current item

#### Test Steps

- [ ] `npm run serve`
- [ ] Visit http://localhost:4200/#/components/markdown-navigator
- [ ] Check it out


src/platform/markdown-navigator/markdown-navigator.component.scss
```diff
@@ -28,4 +28,11 @@
       position: absolute;
     }
   }
+
+  .title {
+    display: inline-block;
+    vertical-align: middle;
+    margin: 8px 0;
+    padding-left: 16px;
```
emoralesb05: I would add the `text-truncate` styles here instead of using the utility class.

src/platform/markdown-navigator/markdown-navigator.component.html
```diff
@@ -13,6 +13,7 @@
         arrow_back
       </mat-icon>
     </button>
+    <span flex *ngIf="currentItemTitle" class="mat-body-2 title text-truncate">{{ currentItemTitle }}</span>
```
emoralesb05: Wondering if we should make this configurable since i can see cases where i dont wanna show a title on a single README?

src/platform/markdown-navigator/markdown-navigator.component.scss
```diff
@@ -28,4 +28,25 @@
...
+
+.truncate {
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.empty-state {
+  padding: 32px;
+  color: #9e9e9e;
```
emoralesb05: This will have to be configurable via theme. For now i would remove the color so its not hardcoded.

PR #1508, opened 2019-10-04
docs(): update new notifications toolbar

emoralesb05:
## Description

Add the new components or functions in the notifications toolbar.

### DIFF
--- a/src/app/components/components/overview/overview.component.ts
+++ b/src/app/components/components/overview/overview.component.ts
@@ -141,6 +141,12 @@ export class ComponentsOverviewComponent {
       route: 'animations',
       title: 'Animations',
     },
+    {
+      color: 'pink-A700',
+      icon: 'build',
+      route: 'functions',
+      title: 'Functions',
+    },
   ];
   optional: Object[] = [
     {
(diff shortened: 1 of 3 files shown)


src/app/components/toolbar/toolbar.component.ts
```diff
@@ -16,7 +16,27 @@ import { getDirection, setDirection } from '../../utilities/direction';
...
+    icon: 'subject',
+    description: 'Render and navigate markdown',
+    title: 'Markdown Navigator'
+  }, {
+    route: '/components/functions',
+    icon: 'build',
+    description: 'New helper functions for multiple situations',
+    title: 'Functions'
+  }, {
+    rout: '/echarts/types/pie',
```
christianmemije: typos

emoralesb05: fixed

src/app/components/components/overview/overview.component.ts
```diff
@@ -161,6 +161,12 @@ export class ComponentsOverviewComponent {
       route: 'flavored-markdown',
       title: 'Flavored Markdown Parser',
     },
+    {
```
christianmemije: Should we add Functions to this page too?

emoralesb05: I think so, since its supposed to be the overview. Ill add it

Issue #1507, opened 2019-10-04
Change dependencies to >= instead of ^


jeremysmartt:
npm install is giving pier dependency errors, need to use >= to get rid of component warnings

christianmemije:
Addressed in #1509

christianmemije:
Oops, #1509 has not been merged in yet.

PR #1509, opened 2019-10-04
feat(): change ^ to ^X.X.X || ^X.X.X on ng versions

emoralesb05:
## Description

Since typically angular support 2 versions back, having ^ throws warnings when moving to a major release change. We will change `^` to `^X.X.X || ^X.X.X` in our package.jsons to be more flexible.

- [ ] `npm run build:lib` still works and see versions applied in `/deploy/platform/`

closes https://github.com/Teradata/covalent/issues/1507


src/platform/core/package.json
```diff
@@ -34,15 +34,15 @@
     "Steven Ov <steven.ov@teradata.com>"
   ],
   "dependencies": {
-    "@angular-devkit/schematics": "^8.1.2",
-    "@angular/cdk": "^0.0.0-MATERIAL"
+    "@angular-devkit/schematics": ">=8.1.2",
```
christianmemije: Do we have to make this change to "@angular-devkit/core in the root level package.json too?

emoralesb05: This is the one i was debating, not entirely sure.

christianmemije:
Can we make the same change to Covalent within the package jsons `^0.0.0-COVALENT` -> `0.0.0-COVALENT`

PR #1538, opened 2019-10-30
feat(): upgrade to angular 9

emoralesb05:
## Description

Upgrade to angular 9 and make sure our build lib process doesnt publish ivy assets.

#### Test Steps

- [ ] `npm install`
- [ ] `npm run serve`
- [ ] Everything component should work fine


src/platform/http/actions/http.mixin.spec.ts
```diff
@@ -4,15 +4,14 @@ import { HttpClientTestingModule, HttpTestingController, TestRequest } from '@an
 import { HttpHeaders, HttpResponse, HttpErrorResponse, HttpParams } from '@angular/common/http';
 import { CovalentHttpModule, TdHttpService } from '@covalent/http';
 import { Observable } from 'rxjs';
-import { TdHttp, TdGET, TdPOST, TdPATCH, TdDELETE, TdParam, TdQueryParams, TdBody } from './';
+import { mixinHttp, TdGET, TdPOST, TdPATCH, TdDELETE, TdParam, TdQueryParams, TdBody } from './';
```
christianmemije: Can we rename `mixinHttp` to something like `TdHttpMixin`?

emoralesb05: We can, but it should be in a separate PR

petebacondarwin:
Great stuff @emoralesb05 and @christianmemije - let me know if you need any further help with Ivy compatibility for v9.

emoralesb05:
Thanks @petebacondarwin ! Gonna merge this so we can start testing it more in our product suite and def will reach out if we encounter any weirdness with the compatibility.

PR #1551, opened 2019-11-06
feat(@covalent/coding-standards): create a @covalent/coding-standards package

christianmemije:
## Description

- Creates a `@covalent/coding-standards` package that we can publish and re-use in other applications in order to pin the configs to packages and package versions.
- Makes husky config re-usable
- Removed rules of stylelint and tslint configs that conflicted with prettier to make extending the configs simpler.

#### Test Steps

- [ ] `npm ci`
- [ ] `npm run build:lib`
- [ ] ensure all files are under `deploy/platform/coding-standards`
- [ ] cd into that dir and do `npm pack`
- [ ] go into another project and do npm install `<path to the tgz file created in command above>`


src/platform/coding-standards/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@covalent/coding-standards",
+  "version": "0.0.0-COVALENT",
+  "description": "Teradata Covalent shared configurations",
```
emoralesb05: Lets use a better description for this.

package.json
```diff
@@ -40,12 +40,6 @@
     "prettier:check": "./node_modules/.bin/prettier --check './**/*.{ts,js,json,css,scss,yml,html}'",
     "commitlint": "node ./node_modules/@commitlint/travis-cli/lib/cli.js"
   },
-  "husky": {
```
emoralesb05: We dont need this?

package.json
```diff
@@ -14,7 +14,7 @@
     "lint": "npm run tslint && npm run stylelint",
```
emoralesb05: Should this run prettier check?

package.json
```diff
@@ -14,7 +14,7 @@
...
     "tslint": "./node_modules/.bin/tslint -p ./tsconfig.json --format codeFrame -c ./tslint.json './src/**/*.ts'",
     "stylelint": "./node_modules/.bin/stylelint 'src/**/*.scss' '!**/_layout.scss' '!**/markdown.component.scss' --config stylelint.config.js --syntax scss",
-    "preinstall": "rm -rf node_modules tmp deploy dist",
+    "preinstall": "rm -rf tmp deploy dist",
```
emoralesb05: We can probably remove this hook.

package.json
christianmemije: We added a `.huskyrc.js` file where we can extend the default husky config inside coding-standards

package.json
christianmemije: Probably, but I would say that is slightly out of scope of this PR

package.json
christianmemije: Removed

src/platform/coding-standards/package.json
christianmemije: Updated

Issue #1502, opened 2019-10-04
Update Covalent Text Editor to use easymde


jeremysmartt:
SimpleMde looks to be a dead project and has vulnerabilities.  This fork called easyMDE looks to be the most active:
https://www.npmjs.com/package/easymde

jeremysmartt:
Steven to Review

PR #1560, opened 2019-11-19
feat(easmde): convert from simplemde

thomaslennan:
## Description
* Converts markdown editor from simplemde to easymde
* Addresses https://github.com/Teradata/covalent/issues/1502

### What's included?
* Substituted easymde for simplemde in package.json
* Replaced all references in code and packaging to easymde.

#### Test Steps
- [ ] `npm run serve`
- [ ] validate markdown editor still works properly in demo


christianmemije:
I would remove `simple` from the description to does not reference simplemde.
Screen Shot 2019-11-19 at 12 31 06 PM

christianmemije:
We should set `autoDownloadFontAwesome` to false. We do not want to be downloading another font.

PR #1602, opened 2020-01-15
Update Utilities Classes with CSS Logical properties

akhedrane:
## Description

Replace utilities classes with [CSS Logical Properties](https://adrianroselli.com/2019/11/css-logical-properties.html) e.g:
push-top     --> push-block-start
pad-right-lg --> pad-inline-end-lg

### What's included?

- Replace utilities classes with CSS Logical properties
- Updated Docs
- Some RTL fixes

#### Test Steps

- [x] `npm run serve`
- [x]  Switch to RTL mode


kyleledbetter:
I can see the potential logic for this request but there’s huge risks:
1. Most importantly this is a massive breaking change forcing large scale search and replace 
2. The proposed classes don’t seem quite as obvious for developers to remember 
3. Is the juice worth the squeeze?

I’d propose to take the approach of angular material and support both (start/end, left/right, etc) and have them consistent

PR #1623, opened 2020-02-11
feat(stylelint): upgrade rules

christianmemije:
### What's included?

- Upgrades stylelint 
- Adds `stylelint-config-standard` rules. Removes some redundant rules. 
- Fixes issues

#### Test Steps

- [ ] `npm run serve`
- [ ] Check everything looks good


package.json
```diff
@@ -13,7 +13,7 @@
     "webdriver-update": "bash ./node_modules/.bin/webdriver-manager update",
     "lint": "npm run tslint && npm run stylelint",
     "tslint": "./node_modules/.bin/tslint -p ./tsconfig.json --format codeFrame -c ./tslint.json './src/**/*.ts'",
-    "stylelint": "./node_modules/.bin/stylelint 'src/**/*.scss' '!**/_layout.scss' '!**/markdown.component.scss' --config stylelint.config.js --syntax scss",
+    "stylelint": "./node_modules/.bin/stylelint 'src/**/*.scss' --config stylelint.config.js --syntax scss --fix",
```
emoralesb05: We should remove the --fix from the check

christianmemije: Removed.

PR #1624, opened 2020-02-13
fix(loading): remove delay on DOM insert/removal

emoralesb05:
## Description

There is an issue with angulars callback animations that introduced in angular@9 when an element that has an animation callback gets removed from the DOM and then added back, the callback stops getting triggered. This made it so that the loading element was not removed from its target after a second insertion.

For this, we decided to remove the animation delay and just remove the loader as soon as the `resolve/removal` action gets triggered. We were already thinking of doing this in the past, so this just gave us yet another reason to do this.

Also included 2 loading demos for replace and observable usage with it.

Note: A unit test was changed since it didnt make sense as it was done in the past.. and it was repurposed for a different case.

#### Test Steps

- [ ] `npm run serve`
- [ ] Go to the loading component demo and check that it works properly

### DIFF
--- a/src/app/content/components/component-demos/loading/demos/loading-demo-observable/loading-demo-observable.component.ts
+++ b/src/app/content/components/component-demos/loading/demos/loading-demo-observable/loading-demo-observable.component.ts
@@ -1,0 +1,21 @@
+import { Component } from '@angular/core';
+
+import { BehaviorSubject, Observable } from 'rxjs';
+
+@Component({
+  selector: 'loading-demo-observable',
+  styleUrls: ['./loading-demo-observable.component.scss'],
+  templateUrl: './loading-demo-observable.component.html',
+})
+export class LoadingDemoObservableComponent {
+  private _subject: BehaviorSubject<any> = new BehaviorSubject<any>({ name: 'I am here!' });
+  observable$: Observable<any> = this._subject.asObservable();
+
+  toggle(): void {
+    this._subject = new BehaviorSubject<any>(undefined);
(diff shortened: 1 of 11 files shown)


src/app/content/components/component-demos/loading/demos/loading-demo-observable/loading-demo-observable.component.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { BehaviorSubject, Observable } from 'rxjs';
+
+@Component({
+  selector: 'loading-demo-observable',
+  styleUrls: ['./loading-demo-observable.component.scss'],
+  templateUrl: './loading-demo-observable.component.html',
+})
+export class LoadingDemoObservableComponent {
+  private _subject: BehaviorSubject<any> = new BehaviorSubject<any>({ name: 'Im here!' });
```
JulieMarie: `Im here!`.....please use proper English.  `'I\'m here!'`

src/app/content/components/component-demos/loading/demos/loading-demo-observable/loading-demo-observable.component.ts
```diff
@@ -0,0 +1,21 @@
...
+  observable$: Observable<any> = this._subject.asObservable();
+
+  toggle(): void {
+    this._subject = new BehaviorSubject<any>(undefined);
+    this.observable$ = this._subject.asObservable();
+    setTimeout(() => {
+      this._subject.next({ name: 'Im here!' });
```
JulieMarie: Same here

PR #1653, opened 2020-02-24
fix(stepper): update demos

JulieMarie:
## Description

Added missing buttons for stepper

### What's included?

- Added missing buttons for stepper

#### Test Steps

- [ ] `npm run serve`
- [ ] Components -> Stepper -> Examples
- [ ] Play around with the examples and verify everything still works

Screen Shot 2020-02-24 at 1 35 04 PM

### DIFF
--- a/src/app/content/components/component-demos/breadcrumbs/demos/breadcrumbs-demo.component.html
+++ b/src/app/content/components/component-demos/breadcrumbs/demos/breadcrumbs-demo.component.html
@@ -1,11 +1,11 @@
-<demo-component [demoId]="'breadcrumbs-demo-basic'">
+<demo-component [demoId]="'breadcrumbs-demo-basic'" [demoTitle]="'Basic Breadcrumbs'">
     <breadcrumbs-demo-basic></breadcrumbs-demo-basic>
 </demo-component>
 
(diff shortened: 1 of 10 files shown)


thomaslennan:
The sample TS has array properties stepsAttrs and stepAttrs. I don't see these used anywhere in the sample HTML. Could these be removed? Or are they required?

There's a slight styling tweak that might be a nit, but if it could be fixed it would be nice. If you are displaying the horizontal oriented stepper and squeeze the browser to minimum width, we're losing a few px of the step titles.

Screen Shot 2020-03-06 at 10 37 57 AM

I don't understand what the pencil implies. Can we add some verbiage around this?

Also, expanding the verbiage at the top of the Stepper page explaining the meaning of active/inactive and completed would be very helpful.

Step three's descriptive text is confusing because where as it was originally in the complete state, you can toggle this state. So really, you're demonstrating the ability to toggle active/inactive in both complete and incomplete states.

I didn't understand by looking at the buttons on the Examples page what Toggle Disable and Toggle First would do. Could the button text be changed to Toggle Disable All and Toggle First Active or something similar?

thomaslennan:
I put together some proposed code changes if that would help. Could do an over the shoulder to see what they look like. If they seem OK, I can push to github.

PR #1654, opened 2020-02-24
feat(nav-steps-demo): vertical stepper example, nomenclature, cleanup

richavyas:
## Description
Add more examples for nav steps

### What's included?
- Vertical nav step
- Added missing routes for horizontal stepper
- Renamed basic component to horizontal

#### Test Steps

- [ ] `npm run serve`
- [ ] then navigate to http://localhost:4200/#/components/nav-steps/examples
- [ ] see horizontal & vertical step navs working

nav-steps


thomaslennan:
Can we please update this example to include at least subtitle text to provide more descriptive information about what each step is demonstrating?

On the examples page, it wasn't clear to me that clicking on the vertical stepper was updating the text at the bottom of the horizontal stepper card. I would recommend putting this same text in each card or pull it out and put it somewhere common (like just below the tabs).

Would be nice if this demo mirrors the stepper demo in that the Overview offered the three tabs to select horizontal, vertical and responsive.

I would recommend we spell out 'and' vs using '&' in the description.

The text "(shrink width...)" does not need to be in () and should be a separate sentence.

Can we explain what the pencil means? I don't think I still understand what that is implying about the state of a step.

PR #1655, opened 2020-02-25
fix(chips): add chips examples

aqupriyanka:
## Description

Add more examples to chips

#### Test Steps

- [ ] `npm run serve`
- [ ] then navigate to https://teradata.github.io/covalent/#/components/chips/overview
- [ ] click on examples section
- [ ] see new examples added

image

image

covalent_chips


thomaslennan:
The directional text "Type here and press ent..." is getting cut off under all circumstances. Truncate to just "Type here"?

Screen Shot 2020-03-06 at 11 28 27 AM

On the examples page, the first example "chips-demo-with-strings" has the "require-match" attribute set. Recommend changing name to "chips-demo-with-requireMatch-attribute".

For the 'chips-demo-for-chip-events' I would recommend adding a span that outputs the most recent event.

thomaslennan:
I put together some proposed code changes if that would help. Could do an over the shoulder to see what they look like. If they seem OK, I can push to github.

PR #1663, opened 2020-03-06
feat(layout): add dense mode

emoralesb05:
## Description

Add dense as an attribute selector so layouts can make their toolbars 48px instead of 64px.

If material ever decides to add a dense option for toolbars, we will just leverage it under the covers.

### What's included?

- `dense` attribute for layouts or added to body

#### Test Steps

- [ ] `npm run serve`
- [ ] add `dense` to either `td-layout-nav`, `td-layout-card-over`, `td-layout-nav-list` or the `body` tag and see toolbars inside the component become 48px height. (this will not affect components outside the component)

### DIFF
--- a/src/platform/core/layout/_layout-theme.scss
+++ b/src/platform/core/layout/_layout-theme.scss
@@ -62,6 +62,28 @@
       z-index: 1;
     }
   }
+
+  body[dense] td-layout-nav,
+  body[dense] td-layout-nav-list,
+  body[dense] td-layout-card-over,
+  td-layout-nav[dense],
+  td-layout-nav-list[dense],
+  td-layout-card-over[dense] {
+    mat-toolbar.td-layout-toolbar {
+      &.mat-toolbar-row,
+      &.mat-toolbar-single-row {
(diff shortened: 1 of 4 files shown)


christianmemije:
What about for the manage list layout?

Screen Shot 2020-03-06 at 11 17 05 AM

emoralesb05:
Manage list doesnt have toolbars, they are injected from a parent component.. i debated applying the style if they are injected.. VS having them add `dense` to the toolbar explicitly.. but we need to add styling for toolbars since that doesnt exist. Thoughts?

christianmemije:
Okay that makes sense. Yeah what about adding a css rule that when there's a toolbar with dense attr, it applies the same style.

emoralesb05:
Well, we need to create a dense style for toolbar in general. Maybe a different PR for that to add into covalent?

christianmemije:
Sounds like a plan. In that case, this is good as is.

emoralesb05:
The concerned will be addressed in https://github.com/Teradata/covalent/pull/1665

PR #1686, opened 2020-03-19
fix(): remove preserveWhitespaces from project

kriswinbush:
## Description

removed preserveWhitespaces key/value from component .ts decorators
### What's included?

- removed all reference to 'preserveWhitespaces: true' from component decorators

#### Test Steps

- [ ] check all component decorators for preserveWhitespaces: true key/value property


src/platform/markdown/markdown.component.spec.ts
```diff
@@ -523,8 +523,7 @@ class TdMarkdownEmptyStaticContentTestRenderingComponent {}
         # title
 
         * list item
-      </td-markdown>`,
-  preserveWhitespaces: true,
+      </td-markdown>`
 })
 class TdMarkdownStaticContentTestRenderingComponent {}
```
christianmemije: We need to make sure that all td-markdown, td-flavored-markdown, td-highlight components have the markup within a string on the ts file. eg move 
```html
<td-markdown>
      {{ markup }}
</td-markdown>
```

emoralesb05: Does it matter for a unit test?

christianmemije: Well I mean we can keep preserve white spaces here since it is already a template string.

PR #1713, opened 2020-03-25
Feat/markdown nav footer

christianmemije:
## Description

Can inject footers into items. Works off of #1712 

#### Test Steps

- [ ] `npm run serve`
- [ ] Verify http://localhost:4200/#/components/markdown-navigator/examples demo works

Screen Shot 2020-03-25 at 15 35 30
Screen Shot 2020-03-25 at 15 35 26

### DIFF
--- a/src/platform/markdown-navigator/markdown-navigator.component.ts
+++ b/src/platform/markdown-navigator/markdown-navigator.component.ts
@@ -20,6 +20,7 @@ export interface IMarkdownNavigatorItem {
   children?: IMarkdownNavigatorItem[];
   description?: string;
   icon?: string;
+  footer?: any;
 }
 
 export interface IMarkdownNavigatorLabels {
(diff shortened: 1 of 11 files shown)


src/platform/markdown-navigator/markdown-navigator.component.ts
```diff
@@ -20,6 +20,7 @@ export interface IMarkdownNavigatorItem {
   children?: IMarkdownNavigatorItem[];
   description?: string;
   icon?: string;
+  footer?: any;
```
emoralesb05: Can you make this type `Type<any>` from angular?

src/platform/markdown-navigator/markdown-navigator.component.ts
```diff
@@ -171,6 +180,12 @@ export class TdMarkdownNavigatorComponent implements OnChanges {
     }
     return undefined;
   }
+
+  get footerComponent(): any {
+    if (this.currentMarkdownItem) {
+      return this.currentMarkdownItem.footer || this.footer;
+    }
```
emoralesb05: Debating if we should show the global even if there is no `currentMarkdownItem`, thoughts?

christianmemije: Tough question, but it does make sense. Because maybe you do want to show a footer at the top level.

christianmemije: Updated PR.

PR #1732, opened 2020-04-13
fix(dialogs): disable user selection for draggable dialogs

aqupriyanka:
## Description
User should not be able select the dialog content while resizing or for draggable dialogs.

### What's included?

- user selection is disabled while dialog resizing.
- user selection is disabled for draggable dialogs.

#### Test Steps

- [ ] `npm run serve`
- [ ] then go to components
- [ ] finally go to dialogs.
- [ ] try any draggable dialog examples.
- [ ] dialog content should not be selected.

### DIFF
--- a/src/platform/core/dialogs/resizable-draggable-dialog/resizable-draggable-dialog.ts
+++ b/src/platform/core/dialogs/resizable-draggable-dialog/resizable-draggable-dialog.ts
@@ -157,6 +157,7 @@ export class ResizableDraggableDialog {
   }
 
   private _handleMouseDown(event: PointerEvent, corner: corners): void {
+    this._renderer2.setStyle(<HTMLElement>this._document.body, 'user-select', 'none');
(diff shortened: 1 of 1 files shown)


src/platform/core/dialogs/services/dialog.service.ts
```diff
@@ -183,6 +183,7 @@ export class TdDialogService {
     matDialogRef.afterOpened().subscribe(() => {
       const dialogElement: HTMLElement = <HTMLElement>this._document.getElementById(matDialogRef.id);
       const draggableElement: DragRef = this._dragDrop.createDrag(dialogElement);
+      this._renderer2.setStyle(dialogElement, 'user-select', 'none');
```
emoralesb05, line 186: When does this get removed? it should only be applied when resizing is happening.

emoralesb05:
image

PR #1735, opened 2020-04-13
fix(): update json-formatter.component.ts

christrude:
## Description
Helping the child parser not do double the work.

Part of an investigation to the formatter not being able to handle large files

### What's included?
- Improved parseChildren function

### DIFF
--- a/src/platform/core/json-formatter/json-formatter.component.ts
+++ b/src/platform/core/json-formatter/json-formatter.component.ts
@@ -212,10 +212,7 @@ export class TdJsonFormatterComponent {
 
   private parseChildren(): void {
     if (this.isObject()) {
-      this._children = [];
-      for (const key of Object.keys(this._data)) {
-        this._children.push(key);
-      }
+      this._children = Object.keys(this._data);
     }
   }
 }


src/platform/core/json-formatter/json-formatter.component.ts
```diff
@@ -212,10 +212,7 @@ export class TdJsonFormatterComponent {
 
   private parseChildren(): void {
     if (this.isObject()) {
-      this._children = [];
-      for (const key of Object.keys(this._data)) {
-        this._children.push(key);
-      }
+      this._children = Object.keys(this._data) ? Object.keys(this._data) : [];
```
christianmemije, line 215: Can't we just do `this._children = Object.keys(this._data);`? An empty object would just return an empty array.

christrude:
> Commit messages are invalid format. I would recommend squashing to fix the first commit message. Also follow https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-guidelines

I thought i amended the commit, thats why i was confused that it still errored. I'll squash instead.

src/platform/core/json-formatter/json-formatter.component.ts
christrude: Yeah just double checked. Can push that.

christianmemije:
@christrude Almost there, just need the commit message format `fix(json-formatter): optimize`

Issue #1736, opened 2020-04-15
Markdown navigator startAt with childrenUrl


christianmemije:
#### Feature Request

When passed in a startAt, be able to also search within childrenUrl. Could do something like `part1.section2.chapter4`. Might require rethinking of startAt search logic.

christianmemije:
Closed in #1739

PR #1739, opened 2020-04-16
feat(markdown-nav): startAt + children_url

christianmemije:
## Description

- Addresses #1736
- `startAt` now type is now `IMarkdownNavigatorItem | IMarkdownNavigatorItem[]` 
-  Can pass in a path as a list to start at, which is useful for a path that depends on  children url

```ts
[{ id: 'test-url-children' }, { id: 'up_up_and_away' }, { id: 'blast_off' }]
``` 

#### Test Steps

- [ ] `npm run serve`
- [ ] See first demo on markdown nav page


christianmemije:
> Any reason you decided on this approach? Just wanna understand your thought process on the array haha.

Yes so originally you had mentioned having a string similar to `id.id.id.id`. I wanted to avoid having to use a certain delimiter that could limit the characters in ids, so I decided to go with an array. Then I decided to go with an array of `IMarkdownNavigatorItems` instead of just ids, since might want to use titles or a diff property to define the path, which would work well with the `compareWith` input.

emoralesb05:
Interesting, makes sense that people might wanna compare with in difference forms.

src/platform/markdown-navigator/markdown-navigator.component.ts
```diff
@@ -367,19 +348,72 @@ export class TdMarkdownNavigatorComponent implements OnChanges {
...
+      }
+      (path || []).forEach((pathItem: IMarkdownNavigatorItem) => this.handleItemSelected(pathItem));
     }
     this._changeDetectorRef.markForCheck();
   }
 
+  private async followPath(
+    items: IMarkdownNavigatorItem[],
+    path: IMarkdownNavigatorItem[],
+    compareWith: IMarkdownNavigatorCompareWith,
```
emoralesb05, line 370: Any reason why you need to pass `compareWith` instead of doing `this.compareWith` inside?

src/app/content/components/component-demos/markdown-navigator/demos/markdown-navigator-demo.component.html
```diff
@@ -1,3 +1,7 @@
+<demo-component [demoId]="'markdown-navigator-demo-children-url-start-at'">
```
emoralesb05, line 1: The code doesnt seem to load for some reason.

emoralesb05: image

src/platform/markdown-navigator/markdown-navigator.component.ts
```diff
@@ -367,19 +348,72 @@ export class TdMarkdownNavigatorComponent implements OnChanges {
...
+  private async _jumpTo(itemOrPath: IMarkdownNavigatorItem | IMarkdownNavigatorItem[]): Promise<void> {
     this.reset();
     if (this.items && this.items.length > 0) {
-      const ancestors: IMarkdownNavigatorItem[] = getAncestors(
-        this.items,
-        item,
-        this.compareWith || defaultCompareWith,
-      );
-      (ancestors || []).forEach((ancestor: IMarkdownNavigatorItem) => this.handleItemSelected(ancestor));
+      const compareWith: IMarkdownNavigatorCompareWith = this.compareWith || defaultCompareWith;
```
emoralesb05, line 354: We can just initialize `this.compareWith` with the `defaultCompareWith` when declaring it and not have to pass it as an argument.

src/platform/markdown-navigator/markdown-navigator.component.ts
```diff
@@ -367,19 +348,72 @@ export class TdMarkdownNavigatorComponent implements OnChanges {
...
+      let path: IMarkdownNavigatorItem[] = [];
+
+      if (Array.isArray(itemOrPath)) {
```
emoralesb05, line 357: Is this assuming the array will always load children by url at every node?

src/platform/markdown-navigator/markdown-navigator.component.ts
christianmemije: Oops, I had moved this outside the class, but moved it back in. Let me remove the param.

src/platform/markdown-navigator/markdown-navigator.component.ts
christianmemije: That assumption is not being made here. Within `followPath()`, we check at each individual node whether there is `children` or a `childrenUrl`

src/platform/markdown-navigator/markdown-navigator.component.ts
christianmemije: Addressed.

src/app/content/components/component-demos/markdown-navigator/demos/markdown-navigator-demo.component.html
christianmemije: Addressed.

src/platform/markdown-navigator/markdown-navigator.component.ts
christianmemije: Addressed. I removed it as a param but still had the || because in case null is passed into compareWith we still want it to fallback to the default.

PR #1742, opened 2020-04-24
fix(coding-standards): fix install

christianmemije:
## Description

There was an issue where the dependencies of `@covalent/commit-standards` were being installed inside of the `@covalent/commit-standards package` (`node_modules/@covalent/commit-standards/node_modules`). Issue explained here: https://stackoverflow.com/questions/14381898/local-dependency-in-package-json#comment36336216_14387210 

A workaround I found was to install it as part of the postinstall script. 

The change in the lock file was due to creating a new one, which was how you could reproduce the issue.

#### Test Steps

- [ ] `npm run serve`
- [ ] `npm ci`
- [ ] npm run prettier:check and verify it runs. 


emoralesb05:
Yeah im ok with this, this is only an issue in covalent because of the "inner" consumption of the package.

PR #1752, opened 2020-05-01
let the user change the theme

youssefbhs8:
i added a button that permits the user to switch between two themes.


christianmemije:
So we don't want Covalent to depend on UI Platform. UI platform already depends on Covalent. UI platform is meant to be a sort of super-set of Covalent, with Teradata specific logic/styling. With that being said, I do think a generic dark theme toggling would be useful. 🤔 Thoughts @philgetz ?

philgetz:
> So we don't want Covalent to depend on UI Platform. UI platform already depends on Covalent. UI platform is meant to be a sort of super-set of Covalent, with Teradata specific logic/styling. With that being said, I do think a generic dark theme toggling would be useful. 🤔 Thoughts @philgetz ?

Agree, Covalent should have its own implementation.

emoralesb05:
I would suggest making the `ui-platform` dark theme a bit more generic and moving it here.

And keep the specifics in `ui-platform` as an `add on` theme

emoralesb05:
Closing this for now as it is not a priority.

PR #1769, opened 2020-06-09
feat(user-menu): handle username and email nicer

christianmemije:
## Description

Small fix to ensure when only a username or email are passed in, it renders correctly.

#### Test Steps

- [ ] `npm run serve`
- [ ] then this
- [ ] finally this

Before
Screen Shot 2020-06-09 at 11 20 50 AM

After
Screen Shot 2020-06-09 at 11 21 02 AM


christianmemije:
Good catch! Sneaky code. Made proper fix.

PR #1799, opened 2020-09-04
Add column sort order for data table

aqupriyanka:
## Description

Add column sort order for data table component
### What's included?

- Add column sort order for data table component

#### Test Steps

- [ ] `npm run serve`
- [ ] then go to components -> Data table
- [ ] check the example with Data table with column sort order

### DIFF
--- a/src/platform/core/data-table/_data-table-theme.scss
+++ b/src/platform/core/data-table/_data-table-theme.scss
@@ -17,6 +17,13 @@
       weight: 600;
     }
   }
+  .td-data-table-sort-badge {
+    font: {
+      family: td-font-family($config);
+      size: td-font-size($config, caption);
+      weight: 400;
+    }
+  }
 }
 
 @mixin td-data-table-theme($theme) {
--- a/src/platform/core/data-table/_data-table-theme.scss
+++ b/src/platform/core/data-table/_data-table-theme.scss
@@ -98,7 +98,7 @@
     color: mat-color($accent);
   }
   .sort-order-bgc {
-    background-color: if($is-dark-theme, map-get($mat-dark-theme-background, background), $white-87-opacity);
+    background-color: if($is-dark-theme, map-get($mat-dark-theme-background, background), mat-color($mat-grey, 200));
     color: if($is-dark-theme, $light-secondary-text, mat-color($mat-grey, 70));
   }
 }
--- a/src/platform/core/data-table/_data-table-theme.scss
+++ b/src/platform/core/data-table/_data-table-theme.scss
@@ -97,7 +97,7 @@
     background-color: mat-color($accent, 0.12);
     color: mat-color($accent);
   }
-  .sort-order-bgc {
+  .td-data-table-sort-badge {
     background-color: if(
       $is-dark-theme,
       map-get($mat-dark-theme-background, background),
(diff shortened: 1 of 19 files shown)


src/platform/core/data-table/data-table.component.html
```diff
@@ -26,22 +26,29 @@
...
-          (mousedown)="_handleStartColumnDrag(i, $event)"
-          (dragstart)="$event?.dataTransfer?.setData('text', '')"
-          (drag)="_handleColumnDrag($event)"
-          (dragend)="_handleEndColumnDrag()"
-          (mouseup)="_handleEndColumnDrag()"
-        >
-          <span class="td-data-table-column-separator"></span>
+        <span [matTooltip]="column.tooltip" [class]="column.columnSortOrder ? 'td-table-header-label' : ''">
+          {{ column.label }}
+          <span *ngIf="column.columnSortOrder" class="push-left-xs header-number-circle mat-caption sort-order-bgc">
```
philgetz, line 34: Sort-order-bgc, push-left-xs and mat-caption should all be covered in the .header-number-circle class, also we should probably rename it to something like td-data-table-sort-badge or something of the "sort".

philgetz (changes requested):
Small css change

src/platform/core/data-table/_data-table-theme.scss
```diff
@@ -17,6 +17,13 @@
       weight: 600;
     }
   }
+  .td-data-table-sort-badge {
+    font: {
+      family: td-font-family($config);
+      size: td-font-size($config, caption);
+      weight: 400;
```
philgetz, line 24: Well "sorted" out! One minor thing here, we can also use the [td-font-weight function from the utilities](https://github.com/Teradata/covalent/blob/fce891282294eec843d40dea6615f623778d723d/src/platform/core/common/styles/_typography-functions.scss#L17) to avoid some hard coding.

src/platform/core/data-table/data-table.component.scss
```diff
@@ -9,8 +9,6 @@ $checkbox-size: 18px;
   justify-content: center;
   align-items: center;
   margin-left: 4px;
-  font-weight: 400;
-  font-size: 12px;
   letter-spacing: normal;
```
philgetz, line 12: I believe we can get rid of this now.

philgetz (changes requested):
Final touch kind of changes!

src/platform/core/data-table/data-table.component.scss
philgetz: I am talking about letter-spacing here

philgetz (approved):
LGTM

src/theme.scss
```diff
@@ -62,6 +62,7 @@ $theme: mat-light-theme($primary, $accent, $warn);
 @include covalent-highlight-theme(); // OR @import '~highlight.js/styles/vs.css';
 @include covalent-flavored-markdown-theme($theme);
 @include teradata-brand($theme);
+@include td-data-table-theme($theme);
```
emoralesb05, line 65: No need to include the theme separately.

src/theme.scss
```diff
@@ -106,6 +107,7 @@ body .mat-card {
   @include covalent-markdown-theme($theme-dark);
   @include covalent-flavored-markdown-theme($theme-dark);
   @include teradata-brand($theme-dark);
+  @include td-data-table-theme($theme-dark);
```
emoralesb05, line 110: Same comment as the previous one.

emoralesb05 (changes requested):

emoralesb05 (approved):

PR #1808, opened 2020-09-17
Feat/angular 10

thomaslennan:
## Description

Upgrade to Angular 10 and address audit issues

### What's included?
- Updated Angular core and material to version 10
- Ran audit fix

#### Test Steps
- [ ] Avoid error trying to update webdriver by setting the following in your environment: export PUPPETEER_SKIP_CHROMIUM_DOWNLOAD='true'
- [ ] `npm run serve`
- [ ] Verify demos still work as expected
- [ ] `npm run serve:prod`
- [ ] Verify demos still work as expected
- [ ] `npm run build:lib`
- [ ] Verify builds to completion


emoralesb05 (changes requested):
I see a lot of changes in the PR that are not pertinent to angular being upgraded.. whats the thought process for that?

package.json
```diff
@@ -66,17 +66,17 @@
...
+    "@angular/common": "^10.1.2",
+    "@angular/compiler": "^10.1.2",
+    "@angular/core": "^10.1.2",
+    "@angular/forms": "^10.1.2",
+    "@angular/material": "^10.2.1",
+    "@angular/platform-browser": "^10.1.2",
+    "@angular/platform-browser-dynamic": "^10.1.2",
+    "@angular/platform-server": "^10.1.2",
+    "@angular/router": "^10.1.2",
     "@ngx-translate/core": "12.0.0",
```
emoralesb05, line 80: We should update the ngx-translate dependencies

https://github.com/ngx-translate/core/releases

src/platform/code-editor/package.json
```diff
@@ -32,7 +32,8 @@
     "Steven Ov <steven.ov@teradata.com>"
   ],
   "dependencies": {
-    "monaco-editor": "0.0.0-MONACO"
+    "monaco-editor": "0.0.0-MONACO",
+    "tslib": "^2.0.0"
```
emoralesb05, line 36: Do we need to add this into every package?

`@angular/material` installs this as a dependency for us.. so not sure we need to do it.

but if we must, lets make it a placeholder so its updated on build and not hardcoded.

src/platform/core/layout/layout-toggle.class.ts
```diff
@@ -19,7 +28,10 @@ export class LayoutToggleBase {}
 /* tslint:disable-next-line */
 export const _TdLayoutToggleMixinBase = mixinDisabled(LayoutToggleBase);
 
-export abstract class LayoutToggle extends _TdLayoutToggleMixinBase implements AfterViewInit, OnDestroy, ICanDisable {
+@Directive()
```
emoralesb05, line 31: Was this a compilation error even for a super class?

src/platform/echarts/base/series/series.component.ts
```diff
@@ -1,23 +1,24 @@
-import { Input, OnChanges, OnInit, OnDestroy } from '@angular/core';
+import { Input, OnChanges, OnInit, OnDestroy, Directive } from '@angular/core';
 
 import { assignDefined } from '../utils/assign-defined';
 import { TdChartOptionsService } from '../chart-options.service';
 import { ITdSeriesTooltip } from '../series/series.interface';
 import { ITdAnimation } from '../base.types';
 
-export interface ITdSeries<T> extends ITdAnimation {
+export interface ITdSeries extends ITdAnimation {
```
emoralesb05, line 8: Do generics fail now?

emoralesb05 (changes requested):

Zwartpet:
Any ETA on this?

src/platform/core/layout/layout-toggle.class.ts
thomaslennan: Angular 10 introduces a new requirement to have decorators on these classes and complained when trying to assign to abstract classes. That's why I removed the generics. The generic aspect didn't seem to be actually in direct usage so I think it's OK to have removed it. It had a big downstream effect on a number of classes that used the base abstract classes, hence the large number of file changes in this PR.

src/platform/code-editor/package.json
thomaslennan: Removed the specific injection of this dependency.

package.json
thomaslennan: Done. Also upgraded the http loader.

src/platform/echarts/base/series/series.component.ts
thomaslennan: Yes. See above. Specifically, usage on abstract base classes pertaining to directives, etc.

emoralesb05 (approved):

craigsmitham:
🎉

PR #1818, opened 2020-10-14
fix(menu): Use default utility class for no padding

9inpachi:
## Description

Use the already existing utility styles class `pad-none` for no padding instead of creating a custom class.

### What's included?

- Using utility class for no padding

#### Test Steps

- [x] `npm run serve`
- [x] Go to http://localhost:4200/#/components/dynamic-menu/examples
- [x] The menu should work like before

None.


emoralesb05:
Nice catch, thanks!

PR #1829, opened 2020-12-14
Feat/angular 11

vhdirk:
## Description
Upgrade to Angular 11

### What's included?
* Updated Angular core and material to version 11

#### Test Steps
- [x] Avoid error trying to update webdriver by setting the following in your environment: 
  `export PUPPETEER_SKIP_CHROMIUM_DOWNLOAD='true'`
- [x] `npm run serve`
  Verify demos still work as expected
- [x] `npm run serve:prod`
  Verify demos still work as expected
- [x] `npm run build:lib`
  Verify builds to completion


Zwartpet:
@jeremysmartt Hi, is the covalent project less supported nowadays? Not so long ago features like this would not stay pending for weeks.

ho0:
@vhdirk I get 4 failed test when I run ng test.

vhdirk:
@ho0 That's odd. I ran `npm run test` (and kept the browser focused). CI also seems to confirm that all tests pass?

ho0:
@vhdirk I ran it 4 times. It passed 3 times and I was almost ready say your right, and then on the 4th time I got 22 errors. So I think we have some kind of unit test dependency issue with the tests. If you run it multiple times do you eventuall get errors?

vhdirk:
@ho0 I've let it run several times now. It's only when I get impatient and forget that this browser window needs constant focus that these tests go sideways. 
I've had a good experience with Jest instead of Jasmine, but seeing how long this PR is open, I'm not expecting substantial changes there either.

vhdirk:
@emoralesb05 @JulieMarie When could we expect a release (beta) that includes this PR? This is currently blocking us from moving to angular 11 for our application.
Thanks for you time and effort!

JulieMarie:
Let me see what I can do @vhdirk to get this done as soon as possible!

PR #1830, opened 2021-02-04
feat(markdown): add youtube video embeds (VAC-442)

blackmjck:
## Description
Page help desperately wants for video capability, so now we can add YouTube embedded videos natively like so:
```![youtu.be/videoID]```

### What's included?
- Native iframe embeds added for youtube share links
- Ability to embed playlists as well as single videos

Screen Shot 2021-02-04 at 2 34 06 PM


fs250044:
Question: Would it be better to have a more generic in order to support potentially more:

![video url]

And then we could potentially support other sources (Vimeo, custom MP4 locations, etc)

PR #1887, opened 2022-02-04
perf(markdown): do not run change detection when DOM timer is fired and scrolled to an anchor

arturovt:


owilliams320:
@arturovt these changes look good 👍🏽 . just need to fix the conflicted files

owilliams320 (changes requested):

owilliams320 (dismissed):
LGTM

PR #1897, opened 2022-03-11
fix(imports): remove require imports from highlight and markdown

owilliams320:
## Description
using ES6 style imports for highlight.js and markdown libraries

### What's included?
- using ES6 style module imports over the node based require
- allows for better module tree shaking

#### Test Steps
- [ ] `npm run start`
- [ ] then check the highlight and markdown demos
- [ ] finally verify it builds

- [ ] `npm run start:prod` still works.

### DIFF
--- a/apps/docs-app/project.json
+++ b/apps/docs-app/project.json
@@ -57,7 +57,7 @@
             {
               "type": "initial",
               "maximumWarning": "500kb",
-              "maximumError": "4mb"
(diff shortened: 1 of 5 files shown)


mrjihai:
Tested it locally.   Works pretty well

mrjihai (approved):
Tested it locally. Works pretty well

PR #1898, opened 2022-03-15
feat(markdown-navigator): adding breadcrumbs

owilliams320:
## Description
Adding TD breadcrumbs to markdown navigator window

### What's included?
- adding breadcrumbs to navigator window per UX mocks
- Fixed live editor demo for markdown navigator window

#### Test Steps
- [ ] `npm run start`
- [ ] then go to localhost:4200
- [ ] finally test the examples in the markdown navigator app

- [ ] `npm run start:prod` still works.

Screen Shot 2022-03-14 at 5 38 56 PM
Screen Shot 2022-03-14 at 5 39 08 PM


adamnel (changes requested):
Have some inconsistencies in how these breadcrumbs are spaced:

- First one is okay
- Second one has extra white space between its the text and the chevron icon
- Third one is misaligned to the other two vertically, probably because it's not in the same container as the text it belongs to like the other two

Screenshot overlay (green is correct, overlay is the current coded version)
image

adamnel (approved):

PR #1904, opened 2022-03-22
fix(release): change up release configuration

owilliams320:
## Description
Modify some of the package versioning for the release step

### What's included?
- only modifies and commits the base package.json on release
- using different replace string for echarts-wordcloud package in libs


.github/workflows/release.yml
```diff
@@ -25,12 +26,8 @@ jobs:
         run: npx nx run-many --target=lint --all
       - name: Test
         run: npx nx run-many --target=test --all
-      - name: Build
-        run: npx nx run-many --target=build --all
```
owilliams320, line 29: build is run as prepare step in the semantic release flow below

mrjihai (approved):

PR #1921, opened 2022-04-01
fix(dynamic-forms): material input name overriding (#1895)

### LINKED ISSUE #1895: Dynamic Forms Input - mat-input bug due to Angular Material 13.1.3+ update - easy fix!
There was a change recently in Angular Material for Material Input that breaks the setting of "name" property on Covalent Dynamic Forms input. Very easy fix by swapping `[attr.name]` with `[name]` in [dynamic forms input](https://github.com/Teradata/covalent/blob/main/src/platform/dynamic-forms/dynamic-elements/dynamic-input/dynamic-input.component.html)

[See issue in Angular Material for more info](https://github.com/angular/components/issues/24463)

tthrash:
Fixes #1895

### DIFF
--- a/libs/angular-dynamic-forms/src/lib/dynamic-elements/dynamic-input/dynamic-input.component.html
+++ b/libs/angular-dynamic-forms/src/lib/dynamic-elements/dynamic-input/dynamic-input.component.html
@@ -8,7 +8,7 @@
       [placeholder]="placeholder"
       [type]="type"
       [required]="required"
-      [attr.name]="name"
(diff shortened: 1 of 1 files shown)


owilliams320 (changes requested):
Looks like there is two other places in dynamic forms where we use matInput and the [attr.name] attribute. Can you please updates those areas also? We can accept this PR after that 👍🏽

owilliams320:
I have time today. will just make the changes in another PR.

PR #1924, opened 2022-04-11
fix(markdown-flavored): teardown subscription when the view is destroyed

arturovt:


owilliams320:
the initial change looks good pls just rebase this branch w/ latest from main

arturovt:
Done. I've only noticed that you removed develop branch but kept pushing to develop.

libs/markdown-flavored/src/lib/flavored-markdown.component.ts
```diff
@@ -98,6 +99,9 @@ export class TdFlavoredMarkdownButtonComponent {
   emitClick(): void {
     this.clicked.emit({ text: this.text, data: this.data });
   }
+  ngOnDestroy(): void {
+    this.clicked.complete();
```
owilliams320, line 103: not sure i 100% support this change. unless clicked event emitter is explicitly called here it's not necessary to call complete on destroy. In my opinion, if a consumer directly subscribed to this event emitter as opposed to using with angulars output templating then it should directly unsubscribe in the parent component

owilliams320 (changes requested):

libs/markdown-flavored/src/lib/flavored-markdown.component.ts
arturovt: Well, I intentionally made this change since unsubscrubing in the parent component becomes a bit complicated. Well, we can add `takeUntil` to `clicked` before subscribing within the `_replaceButtons` method. What happens if the `TdFlavoredMarkdownButtonComponent` is destroyed before the parent component is destroyed? The subscribtion would exist until the parent component is destroyed. That's why I thought to call `.complete()` on the subject and added comments there since the parent and the child have different life spans.

owilliams320: would prefer to use takeUntil where it subscribes below.

arturovt: @owilliams320 hey, updated, added takeUntil to where it subscribes.

owilliams320 (approved):
LGTM

PR #1929, opened 2022-04-11
perf(code-editor): do not run change detection whenever the editor does re-layout

arturovt:


libs/angular-code-editor/src/lib/code-editor.component.ts
```diff
@@ -352,32 +362,29 @@ export class TdCodeEditorComponent
...
+        this._heightSubject.asObservable().pipe(distinctUntilChanged())
+      )
+        .pipe(debounceTime(100), takeUntil(this._destroy))
+        .subscribe(() => {
+          // Note: this is being called outside of the Angular zone since we don't have to
+          // run change detection whenever the editor resizes itself.
+          this.layout();
+        })
+    );
+    this._ngZone.runOutsideAngular(() =>
```
owilliams320, line 378: not sure this really needs to be run outside of the ngzone? not seeing what would trigger change detection if its just getting values and triggering next in a behavior subject

owilliams320 (changes requested):

libs/angular-code-editor/src/lib/code-editor.component.ts
arturovt: There's _timer_ (that acts as a setInterval) that triggers change detection

arturovt:
@owilliams320 friendly ping

owilliams320 (approved):

PR #1930, opened 2022-04-11
ci: setup ESLint RxJS plugins

arturovt:
There're no functional changes in this PR. It only setups RxJS plugins to ensure that existing/subsequent observables are unsubscribed by developers. I've also replaced `subscription = obs.subscribe` with `obs.pipe(takeUntil()).subscribe` in some places to be consistent with the other code.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,7 +1,7 @@
 {
   "root": true,
   "ignorePatterns": ["**/*"],
-  "plugins": ["@nrwl/nx"],
+  "plugins": ["@nrwl/nx", "rxjs", "rxjs-angular"],
   "overrides": [
     {
       "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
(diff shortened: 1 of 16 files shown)


libs/markdown-flavored/src/lib/flavored-markdown.component.ts
```diff
@@ -130,6 +130,7 @@ export type IReplacerFunc<T> = (
   templateUrl: './flavored-markdown.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
+// eslint-disable-next-line rxjs-angular/prefer-takeuntil
```
owilliams320, line 133: does this need to be here?

.eslintrc.json
```diff
@@ -29,7 +29,14 @@
...
       "files": ["*.ts", "*.tsx"],
       "extends": ["plugin:@nrwl/nx/typescript"],
-      "rules": {}
+      "parserOptions": {
+        "project": ["tsconfig.*?.json"],
+        "warnOnUnsupportedTypeScriptVersion": false
+      },
+      "rules": {
+        "rxjs/no-unsafe-takeuntil": "error",
+        "rxjs-angular/prefer-takeuntil": "error"
```
owilliams320, line 38: seeing as there is a few spots here that disable this rule. my vote would be to either make this a warning and remove the spots where this rule gets disabled....or not include this rule at all

owilliams320 (changes requested):

owilliams320 (dismissed):

owilliams320 (approved):

PR #1934, opened 2022-05-11
fix(dynamic-forms): fixed defualt value for booleans

### LINKED ISSUE #1876: Bug - Boolean default value in dynamic select component doesn't appear if false
Bug:
In the project my team has been working on we have found an edge case when using the dynamic element select component. We wanted to create a custom component for boolean values rather than using the built-in "TdDynamicType" of boolean, we elected to use the "TdDynamicElement" Select. By using select we wanted to have two selections which would have the values true and false presented as "Yes" and "No" to the user. We saw that if the value of the element defaulted to true the selection "Yes" was selected for the user. If it was defaulted to false then nothing was shown. After a brief look at the code the below seems to be the culprit,

https://github.com/Teradata/covalent/blob/f983e55a99b32e0e97d94b028d9b7bfcf4207db0/src/platform/dynamic-forms/dynamic-elements/dynamic-select/dynamic-select.component.html#L11

Our selection object is as follows,
```
element.selections = []
element.selections.push({
  label: 'Yes',
  value: true
} as any)
element.selections.push({
  label: 'No',
  value: false
} as any)
...
...
element.default = false
```
As you can see if false is chosen the value will not be shown. 

It would be nice if you could use boolean values in the select component.

owilliams320:
## Description

closes #1876

setting the default value with a boolean of `false` will not correctly set the default value.

- [ ] `npm run lint` passes.

### DIFF
--- a/libs/angular-dynamic-forms/src/lib/dynamic-elements/dynamic-select/dynamic-select.component.html
+++ b/libs/angular-dynamic-forms/src/lib/dynamic-elements/dynamic-select/dynamic-select.component.html
@@ -10,7 +10,7 @@
     >
       <mat-option
         *ngFor="let selection of selections"
-        [value]="selection.value || selection"
+        [value]="selection.value ?? selection"
       >
         {{ selection.label || selection }}
       </mat-option>


owilliams320:
definitely. looks like there is not much coverage at all on in this package

PR #1936, opened 2022-05-13
feat(markdown-navigator): fix added new window icon for help component

sitap1312:
## Description
(VAC-199)Markdown navigator - Help component - Added new window icon for help.

### What's included?

- One
- Two
- Three

#### Test Steps

- [x] `npm run start`
- [ ] then this
- [ ] finally this

- [x] `npm run lint` passes.

New-window-Icon


owilliams320 (changes requested):
whats is this button supposed to do? how would a consumer of this component use this? also  we should not always be showing this button. perhaps an input to hide/show on the component.

owilliams320:
the button should also be a using button icon. https://material.angular.io/components/button/examples

PR #1943, opened 2022-05-18
fix:adding-dock-icon

Msubhashini1:
Implement docked mode VAC - 1332

Added Dock Icons and implemented code in the help navigator. 
after clicking on the undock Icon it will replace with the dock_right Icon (vice versa)

### What's included?
added dock Icons in the help navigator.

- One
- Two
- Three

#### Test Steps
Description : Implemented minimized mode for the help component.
Test Steps: 1. Log in into the console app.
2. From left menu click on help icon, (i.e) last Icon from bottom.
3. Help window will be open.
4. Inside window there will be undock icon.
5. After we clicking on the undock icon it will replace with the dock_right icon (vise versa).

- [x] `npm run start`
- [ ] then this
- [ ] finally this

- [x] `npm run lint` passes.

undock 2022-05-18 at 4 24 31 PM

dock_right 2022-05-18 at 4 24 46 PM


christrude:
Why does undockState just switch the icon, instead of switching it to a docked view. Will that happen in it's use instead? I'm not sure this is the solution.

adamnel:
There should be no gap between the icon buttons (the close icon and the dock icon). Right now looks like the markup has a space/line return in it which is separating them more than they should be.

Also, please verify in dark theme if you haven't already.

PR #1944, opened 2022-05-19
fix(flavored-markdown): allow for all list characters

owilliams320:
## Description

Adjusted the regex for parsing lists in the flavored markdown component

### What's included?
- flavored markdown parsing for lists now include `+,*` where before it was just `+`

#### Test Steps
- [ ] `npm run start`
- [ ] then go to the flavored markdown examples
- [ ] finally in code change the example lists for flavored markdown to use the three different supported list characters

- [ ] `npm run lint` passes.


JulieMarie (dismissed):
LGTM

JulieMarie (approved):
LGTM

PR #1947, opened 2022-05-23
fix(rxjs): in package from ~ to ^

JulieMarie:
## Description

Changed versioning of rxjs in package.json from `~7.4.0` to `^7.4.0`

#### Test Steps

- [ ] `npm run start`

- [ ] `npm run lint` passes.


owilliams320 (approved):
LGTM

PR #1957, opened 2022-07-26
perf(file): do not run change detection on `click` and `keyup` events when the `td-file-input` button is clicked

arturovt:


owilliams320 (approved):
LGTM

PR #1975, opened 2022-09-29
fix(componets): deploy to npm & chromatic ci

owilliams320:
## Description

Fixing the build stage for the components library

### What's included?

- @covalent/components will now build with the rest of the libraries to also be published
- fixing the chromatic workflow to only run when certain libs change (components, tokens, icons)
- also using npx cli command for chromatic instead of action since its not verified by github marketplace

#### Test Steps

- [ ] `nx build components`

- [ ] `npm run lint` passes.


mikespudwebb (approved):
Awesome sauce!

PR #1982, opened 2022-11-02
feat(components): adding sidesheet component

owilliams320:
## Description

Adding side sheet component to the web component library

### What's included?

- Added sidesheet component to the components library
- Added storybook for side sheet component

#### Test Steps

- [ ] `npm run storybook`
- [ ] then log in an scroll to the side sheet story in the side bar
- [ ] finally test the side sheet component

- [ ] `npm run lint` passes.

sidesheet

### DIFF
--- a/libs/components/src/side-sheet/side-sheet.scss
+++ b/libs/components/src/side-sheet/side-sheet.scss
@@ -1,5 +1,5 @@
 :host {
-  --td-side-sheet-width: 400px;
+  --td-side-sheet-width: 800px;
 }
 
 .mdc-dialog {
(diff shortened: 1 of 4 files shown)


libs/components/src/side-sheet/side-sheet.scss
```diff
@@ -0,0 +1,48 @@
+:host {
+  --td-side-sheet-width: 400px;
```
adamnel, line 2: Let's set the default here to 800

adamnel (dismissed):

adamnel (approved):

PR #1984, opened 2022-11-04
fix(flavored-markdown): list display fixes, CRLF fix

DersJ:
## Description

Fixing an issue with markdown navigator which doesn't display links within lists if the links contain a `+` in their query params.

### What's included?

- Adding back `-` as a list character for CFM lists
- Making CFM lists opt-in via `useCfmList` input parameter. By default vanilla list will be used.
     - using cfm within readme viewer for docs app
- Fix for #1987 - removing `/r` characters from markdown before parsing/conversion

#### Test Steps

- [ ] `npm run start`
- [ ] then this
- [ ] finally this

- [ ] `npm run lint` passes.

Broken:

Screen Shot 2022-11-07 at 6 30 39 PM

Fixed:
Screen Shot 2022-11-07 at 6 28 53 PM


mikespudwebb (dismissed):
I can't think of why this would be bad. 🤷‍♂️

christrude (approved):

PR #1986, opened 2022-11-04
Fix/components various fixes

owilliams320:
## Description

Adding a few various fixes after testing out implementation of compoments for a prototype

### What's included?

- Fixes card component not being built in package
- Added icon button toggle component missed from initial commit
- Fixed the accent color not workign for just plain buttons (now works for all variations)
- Removed the boiler plate breadcrumb code from the toolbar component (now is just a small extension of top-app-toolbar)

#### Test Steps

- [ ] `npm run storybook`
- [ ] then this open brower window where the storybook instance opened at
- [ ] finally this test the additions above in their stories (toolbar, icon-button-toggle, button)

- [ ] `npm run lint` passes.


libs/components/src/button/_button.theme.scss
```diff
@@ -12,20 +12,13 @@
...
     --mdc-button-outline-color: var(--mdc-theme-primary);
   }
 
   // --- apply error color --- //
-  td-button[outlined].destructive.primary,
-  td-button[raised].destructive.primary,
-  td-button[text].destructive.primary,
-  td-button[outlined].negative.primary, // Same thing, just using "negative" alias
-  td-button[raised].negative.primary,
-  td-button[text].negative.primary {
```
mikespudwebb, line 28: Do we want the `negative.primary` pulled over too?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "covalent",
-  "version": "4.10.1",
+  "version": "4.10.0",
```
mikespudwebb, line 3: Weird to go "backwards" with a version number! 🤪

mikespudwebb (approved):
Couple of questions, but great cleanup otherwise.

mikespudwebb (approved):

PR #1995, opened 2022-11-29
feat(components): adding code snippet component

owilliams320:
## Description

Adding Code snippet web component to the web component library

Figma (needs td login)
https://www.figma.com/file/4qn8W9GCuhyNxEMBhMeahz/Code-samples?node-id=0%3A1&t=Rs15lTIuTGzTzrxQ-1

### What's included?

- Adding code snippet using [highlight.js](https://highlightjs.org/)
- Supports the common [34](https://highlightjs.org/static/demo) languages 
- Uses atom one light and dark theme
- Supports adding a title in the header and a slot for action

#### Test Steps

- [ ] `npx run components:storybook`
- [ ] then wait for the browser window open
- [ ] finally navigate to the code snipppets story

- [ ] `npm run lint` passes.

code-snippet

### DIFF
--- a/libs/components/src/code-snippet/code-snippet.scss
+++ b/libs/components/src/code-snippet/code-snippet.scss
@@ -1,0 +1,74 @@
+:host {
+  background-color: var(--mdc-theme-surface-canvas);
+  border-radius: var(--mdc-shape-medium);
+  display: block;
+  position: relative;
+
+  /* For max height*/
+  overflow: hidden;
+
+  pre {
+    margin: 0;
+    padding: 16px 0;
+    overflow: auto;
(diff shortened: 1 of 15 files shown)


adamnel:
@owilliams320 The default theme should match that of the parent app, but allow you to use the icon button to reverse it. So in this demo it should start as dark themed.

owilliams320:
> @owilliams320 The default theme should match that of the parent app, but allow you to use the icon button to reverse it. So in this demo it should start as dark themed.

For this i added the code snippet tokens to the to `tokens` package so we can use as variables to switch between like the rest of the system. 
theme-toggle-cod

adamnel (approved):

PR #1998, opened 2022-12-05
feat(components): adding text lockup component

owilliams320:
## Description

Adding text lock up component to display basic key value text anywhere

_**Using visuals pulled from this [figma](https://www.figma.com/file/ltYM4aeDULd3ec6Sd8Xvpk/Typography?node-id=2233%3A1034) (not official) **_
### What's included?

- Text lockup component with support for small and large scale
- Icon support for subtext
- State support for `active, positive, negative, caution`
- Trailing subtext support for to see the subtext on the bottom of the main text 

#### Test Steps

- [ ] `npx nx run components:storybook`
- [ ] then locate the text lockup story under Components -> Text lockup
- [ ] finally test the different storys and parameters configured

- [ ] `npm run lint` passes.

text-lockup

### DIFF
--- a/libs/components/src/index.ts
+++ b/libs/components/src/index.ts
@@ -34,6 +34,7 @@ export * from './tab/tab';
 export * from './tab/tab-bar';
 export * from './textarea/textarea';
 export * from './textfield/textfield';
+export * from './text-lockup/text-lockup';
 export * from './toolbar/toolbar';
 export * from './top-app-bar/top-app-bar';
 export * from './top-app-bar/top-app-bar-fixed';
--- a/libs/components/src/text-lockup/text-lockup.scss
+++ b/libs/components/src/text-lockup/text-lockup.scss
@@ -1,0 +1,66 @@
+.subtext--trailing {
+  display: flex;
+  flex-direction: column-reverse;
+}
+
+.subtext {
+  display: flex;
+  align-items: center;
+
+  --mdc-icon-size: 16px;
+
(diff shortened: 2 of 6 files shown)


adamnel:
So looking at this now I think we have two separate atomic components on our hands here:

### Text + text lockup
image

### Icon + text lockup
For the icon + text lockup, we'd want one for each step on the type scale, something like this:

image

Then if you wanted to do what you have in your screenshots where there's an icon in the label of the Text + Text lockup you just nest that icon + text lockup into that slot.

I do think they're different though, for example the Icon + text lockup belongs in various places like data tables, lists, maybe sometimes even inline in body copy:

image

And if it's locked away in the larger text + text lockup component that precludes that doesn't it?

owilliams320:
> ### Icon + text lockup



> So looking at this now I think we have two separate atomic components on our hands here:
> 
> ### Text + text lockup
> image
> 
> ### Icon + text lockup
> For the icon + text lockup, we'd want one for each step on the type scale, something like this:
> 
> image
> 
> Then if you wanted to do what you have in your screenshots where there's an icon in the label of the Text + Text lockup you just nest that icon + text lockup into that slot.
> 
> I do think they're different though, for example the Icon + text lockup belongs in various places like data tables, lists, maybe sometimes even inline in body copy:
> 
> image
> 
> And if it's locked away in the larger text + text lockup component that precludes that doesn't it?

I agree we can probably benefit from adding either a seperate component with `td-icon-text` with support for status tokens. or adding icon support to the `td-typography` component we have today

PR #2014, opened 2023-03-13
fix(components): adjusted import location of scss theme file

owilliams320:
## Description

Fixing the import path of the theme file for the components storybook preview. 

Note: see the broken css here https://teradata.github.io/covalent/docs/components/

#### Test Steps

- [ ] `npm run storybook`
- [ ] then wait for the browser window to pop open
- [ ] then observe that css now looks correct

- [ ] `npm run lint` passes.


mikespudwebb (approved):
Looks good to me!

PR #2030, opened 2023-06-14
fix(components): adding secondary entry points for modules

owilliams320:
## Description

Adding secondary entry points for covalent/components package fore each module. For example importing the button should be possible now by using "@covalent/button" path in either an ESM or CommonJS format

### What's included?

- Added secondary entry point mapping in package.json
- Added default export for each module
- Removed the suffix`Base` to simplify some of the modue class names

#### Test Steps

- [ ] `npx nx run components build`
- [ ] `npx nx run components test`
- [ ] `npm run storybook`
- [ ] then open the browser
- [ ] finally test all the components operate the same

- [ ] `npm run lint` passes.


mikespudwebb (approved):
Love this!

PR #2070, opened 2023-11-06
fix(markdown): external icon for target _blank

### LINKED ISSUE #2063: External links should use the external link icons, by default
**Is your feature request related to a problem? Please describe.**
When viewing a MD text hyperlink, users have no way to know if the link will open an external browser window or not.

**Describe the solution you'd like**
The **open in new** icon should be added to links that open a new window (that is, that link to something outside of the current domain).

**Describe alternatives you've considered**
The image/icon could manually be added to every link, but that (1) becomes a maintenance nightmare and (2) makes it impossible to change out iconsets.

**Additional context**
Would be nice to be able to:
1. Define the exact icon to use (e.g., **https://fonts.google.com/icons?icon.query=open+in+new**)
2. Define the exact "home" domain.
3. Be able to also add (or not) the icon on-demand.

owilliams320:
## Description

closes #2063 

Adding external link icon for links in markdown parsers that use `target="_blank"`

#### Test Steps

- [ ] `npm run start`
- [ ] then open the docs in browser
- [ ] finally look at any documentation file that external links and verify the icon is there

- [ ] `npm run lint` passes.

Screenshot 2023-11-06 at 11 01 06 AM

### DIFF
--- a/libs/markdown/src/lib/markdown.component.scss
+++ b/libs/markdown/src/lib/markdown.component.scss
@@ -5,6 +5,21 @@
       background-color: transparent;
     }
 
+    a[target='_blank'] {
+      display: inline-flex;
+      align-items: center;
+
+      &::after {
+        content: 'open_in_new';
+        font-family: 'Material Icons', sans-serif;
+        padding-left: 2px;
+
+        &:hover {
+          text-decoration: none;
+        }
+      }
+    }
+
     a:active,
     a:hover {
       outline-width: 0;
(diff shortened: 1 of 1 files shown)


adamnel:
Need to remove the underline from the icon on hover:

image

Otherwise looks good - text is wrapping above and below without the icon pushing the line-height wider on rows where it's present, which was my biggest concern.

adamnel (approved):

PR #2104, opened 2024-02-22
build(echarts): modifying echarts stats package version placeholder

owilliams320:
## Description

Adjusting the placeholder values for ECharts stat package. This value was not being correctly replaced prior to being pushed to NPM


blackmjck (approved):
Looks good to me!

PR #2116, opened 2024-03-08
fix: remove token generation scripts, workflows and related files

bsahitya:
## Description

Remove token generation scripts, files and workflows which are no longer needed. 

### What's included?

- Deleted `_generated` folder from token library
- Deleted token generation script and github workflow
- Removed token generation command from package.json

- [x] `npm run lint` passes.


owilliams320 (approved):
LGTM. thanks for this cleanup @bsahitya

PR #2122, opened 2024-03-14
fix(icons): fix path exports

owilliams320:
## Description

fix for exports not picking up the correct css file for tokens package


bsahitya (approved):
Looks good to me!

PR #2123, opened 2024-03-14
docs(knapsack): modifying design tokens path

owilliams320:
## Description

- Moving the design tokens path to have a `cv` prefix

Screenshot 2024-03-14 at 5 20 52 PM


bsahitya (approved):
Looks good to me!

PR #2133, opened 2024-03-25
fix(tokens): remove % from scss variable names

bsahitya:
## Description

SCSS tokens which end with `%` are not showing up in the DOM.

### What's included?

- Remove the trailing `%` symbol in scss variables

#### Test Steps

- [x] `npm run start`
- [x] Access any of the css variables like (--cv-theme-on-surface-16) in the DOM

- [x] `npm run lint` passes.

##### Screenshots
Screenshot 2024-03-25 at 1 40 40 PM

### DIFF
--- a/libs/tokens/src/color/covalent-m3.json
+++ b/libs/tokens/src/color/covalent-m3.json
@@ -2717,7 +2717,7 @@
             }
           }
         },
-        "primary-8%": {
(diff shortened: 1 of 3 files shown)


owilliams320 (approved):
LGTM 👍🏽

PR #2153, opened 2024-04-29
feat(components): moving subnav property to nav-item component

owilliams320:
## Description

moving the required `subNav` property from the list component scss to the `cv-navigation-item` component. This makes the styling for the subnav items to be more robust when wrapped with an other elements like an `a` tag

#### Test Steps

- [ ] `nx serve component`
- [ ] then open the localhost
- [ ] finally test that visually nothing has changed

- [ ] `npm run lint` passes.


bsahitya (approved):

bsahitya (approved):
Looks good!

PR #2154, opened 2024-04-30
Feat/ng standalone component update

fernando182gomez:
## Description

Upgrade angular standalone components

- [ ] `npm run lint` passes.


owilliams320:
@fernando182gomez can you add a list of which components are now standalone, the modules that will be deprecated and which components are still needing to be done?

PR #2163, opened 2024-05-21
fix(app-shell): small fix to works fine the resizable

vicalcantFrontEnd:
## Description

Small fixes to works fine the Help Resizable  property..

### What's included?

- change the variable helpResizable instead resizing into the app-shell.stories.js

#### Test Steps

- [ ] `npm run start`
- [ ] then this
- [ ] finally this

- [ ] `npm run lint` passes.

app-shell-fixes

### DIFF
--- a/libs/components/src/app-shell/app-shell.ts
+++ b/libs/components/src/app-shell/app-shell.ts
@@ -156,7 +156,7 @@ export class CovalentAppShell extends DrawerBase {
       this._startWidth = this.helpWidth;
       document.addEventListener('mousemove', this._resize);
       document.addEventListener('mouseup', this._stopResize);
-      (event.target as HTMLElement).classList.add('resizing');
+      (event.target as HTMLElement).classList.add('helpResizable');
     }
 
     resizeHandle?.addEventListener('dblclick', () => {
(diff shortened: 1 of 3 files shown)


libs/components/src/app-shell/app-shell.ts
```diff
@@ -156,7 +156,7 @@ export class CovalentAppShell extends DrawerBase {
       this._startWidth = this.helpWidth;
       document.addEventListener('mousemove', this._resize);
       document.addEventListener('mouseup', this._stopResize);
-      (event.target as HTMLElement).classList.add('resizing');
+      (event.target as HTMLElement).classList.add('helpResizable');
```
bsahitya, line 159: I think if the class name is changed here, it should be changed in the scss file too.
There a couple of styles being applied to this class.

bsahitya (changes requested):

libs/components/src/app-shell/app-shell.ts
vicalcantFrontEnd: make sense... already pushed these changes

owilliams320 (approved):

bsahitya (approved):

PR #2165, opened 2024-05-29
Feat/status dialog

bsahitya:
## Description

- Add status dialog component
- Add knapsack documentation for the component

### What's included?

- A status dialog component with the states: error|warning|positive
- Allows uses to optionally add additional details
- Knapsack documentation for the components

#### Test Steps

- [x] `npm run storybook`
- [x] then open status dialog
- [x] view the demos and change the controls to interact with the component

- [x] `npm run lint` passes.

Screenshot 2024-05-29 at 3 16 35 PM
Screenshot 2024-05-29 at 3 16 45 PM
Screenshot 2024-05-29 at 3 16 57 PM
Screenshot 2024-05-29 at 3 17 09 PM
Screenshot 2024-05-29 at 3 17 43 PM


owilliams320 (dismissed):
Great work!

adamnel:
image

Dunno if it's a local problem for me but the colored bar to the left side isn't showing up. 
image

bsahitya:
> image
> Dunno if it's a local problem for me but the colored bar to the left side isn't showing up. image



> image
> Dunno if it's a local problem for me but the colored bar to the left side isn't showing up. image

@adamnel This is an issue with the covalent scss tokens not being converted correctly to css variables. The `%` symbol in some of the scss tokens is causing the issue. The fix is included in this PR. So once the PR is merged all the tokens will be available and the colors will be visible. But, to test locally, you would have to build the tokens again and include them in the project.
- Run the command `npx nx run tokens:build`
- This will create a dist folder in the project root.
- Navigate to dist/libs/tokens copy all the files.
- Navigate to node_modules/@covalent/tokens and replace the files with the newly copied files.

owilliams320 (dismissed):

adamnel (approved):
Tweaked positioning of close icon, check my work

RDUBPhoto (approved):

libs/components/src/status-dialog/status-dialog.scss
```diff
@@ -2,13 +2,17 @@
   --mdc-dialog-content-ink-color: var(--cv-theme-on-surface-variant);
   --mdc-dialog-heading-ink-color: var(--cv-theme-on-surface);
   --mdc-dialog-min-width: 96vw;
-  --mdc-shape-medium: 12px;
+  --mdc-shape-medium: calc((var(--mdc-shape-small)) * 1.5);
```
adamnel, line 5: This should just respect whatever the `--mdc-shape-medium` value is. If we want to change that we'd do it globally. Is this missing or something and that's why you're having to calculate it?

bsahitya: Globally this value is set to 8px. Components like code snippet use this value for their border radius. But the design spec provided for error messaging has a border radius of 12px. So I calculated it again. If not we will have to create a new covalent token just for the border radius of this status dialog.

bsahitya: Ok. For now, I removed the override in this component. Will check with the team and set the global medium size to 12 after analysing which components it will affect.

owilliams320 (approved):

PR #2174, opened 2024-06-13
Data management icon

carolineguzylak:
## Description

Updated data management icon and added outlined version

### What's included?

Updated data management icon and added outlined version
#### Test Steps

- `npm run storybook`
- View icons under covalent icons

- [ ] `npm run lint` passes.


adamnel:
This is quite a lot of files updated to change one icon isn't it?

carolineguzylak:
> This is quite a lot of files updated to change one icon isn't it?

Yep, cause the font file has to be regenerated and then the new icon variable has to be added in 3 different files...

PR #2180, opened 2024-07-09
fix(markdown): allow relative paths

JulieMarie:
## Description

Now markdown will work with both Absolute and Relative paths to allow for offline capabilities

### What's included?

- Accept Relative paths as well as Absolute paths

#### Test Steps

- [ ] `npm run start`
- [ ] then this
- [ ] finally this

- [ ] `npm run lint` passes.


JulieMarie:
Create PR against wrong branch, closing this PR.

PR #2184, opened 2024-07-12
fix(components): list item disabled

owilliams320:
## Description

Adding style to a list item with a disable property added

### What's included?

- styling in the base list component to set the color to us `--cv-theme-on-surface-38` when disabled
- icons use `--cv-theme-on-primary-container-16` token when disabled
- adjusted list stories in storybook

#### Test Steps

- [ ] `npm run storybook`
- [ ] then navigate to the list item story
- [ ] finally select `disabled` under the story controls to see the changes

- [ ] `npm run lint` passes.

Screenshot 2024-07-12 at 12 28 37 PM
Screenshot 2024-07-12 at 12 49 18 PM

### DIFF
--- a/libs/components/src/list/list-item.scss
+++ b/libs/components/src/list/list-item.scss
@@ -13,6 +13,13 @@
 
 :host([disabled]) {
   color: var(--cv-theme-on-surface-38);
+
+  &:host([graphic='avatar']) .mdc-deprecated-list-item__graphic,
+  .mdc-deprecated-list-item__meta {
+    --mdc-theme-text-icon-on-background: var(
+      --cv-theme-on-primary-container-16
+    );
+  }
 }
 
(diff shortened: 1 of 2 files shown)


libs/components/src/list/list-item.scss
```diff
@@ -11,6 +11,17 @@
   border-radius: 9999px;
 }
 
+:host([disabled]) {
+  color: var(--cv-theme-on-surface-38);
+
+  &:host([graphic='avatar']) .mdc-deprecated-list-item__graphic,
+  .mdc-deprecated-list-item__meta {
+    --mdc-theme-text-icon-on-background: var(
+      --cv-theme-on-primary-container-16
```
adamnel, line 20: This should also be `--cv-theme-on-surface-38`

adamnel (changes requested):
One color tweak

PR #2198, opened 2024-07-22
fix(components): styling adjustment to cards web component

owilliams320:
## Description

styling fix for cards that are non-interactive

### What's included?

- using primary card class only when card has ripple affect and is primary action only

#### Test Steps

- [ ] `npm run storybbok`
- [ ] then navigate to the card story
- [ ] finally check make sure everything is ok

- [ ] `npm run lint` passes.


JulieMarie (approved):
LGTM

PR #2205, opened 2024-07-24
fix(): use is-dark for highligh dark theme (#2053)

JulieMarie:
## Description

Missed a commit that needed to be on 4.x.x

#### Test Steps

- [ ] `npm run start`
- [ ] then this
- [ ] finally this

- [ ] `npm run lint` passes.


owilliams320 (approved):
LGMT

PR #2206, opened 2024-07-25
Fix/app shell frame tweaks

adamnel:
## Description

General theme tweaks, mostly changing base surface color from pinkish to blueish.

### What's included?

- Change surface and foreground colors to a more blue tint (as opposed to the pinkish previously)
- Change default color of non-selected nav items
- Change default color of non-header body text
- Remove extra tracking from button typography

#### Test Steps

- [ ] `npm run storybook`
- [ ] then navigate to the app shell story
- [ ] compare against live product 

- [ ] `npm run lint` passes.

##### Before
app-shell-frame-tweaks_before

##### After
app-shell-frame-tweaks_after

### DIFF
--- a/libs/components/src/app-shell/app-shell.stories.js
+++ b/libs/components/src/app-shell/app-shell.stories.js
@@ -105,7 +105,7 @@ const Template = ({
     ${contained ? `contained` : ''}
     ${fullWidth ? `fullWidth` : ''}
     ${helpResizable ? `helpResizable` : ''}
-    
+
     >
 
       <cv-icon-button slot="section-action" icon="arrow_back"></cv-icon-button>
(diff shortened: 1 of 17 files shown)


libs/components/src/app-shell/app-shell.stories.js
```diff
@@ -188,11 +188,11 @@ const Template = ({
           ></cv-icon-button>
           <cv-icon-button
             slot="actionItems"
-            icon="close"
+            icon="close"I
```
owilliams320, line 191: Looks like this "I" is by mistake

owilliams320 (changes requested):
Looks good overall just have a couple of questions

owilliams320:
Can we get some screen shots please

PR #2218, opened 2024-08-05
Feat/fullscreen takeover

bsahitya:
## Description

Add a new web component called `Fullscreen Takeover`.
It uses the `cv-dialog` component internally.

### What's included?

- Fullscreen takeover component with a default slot for content.
- Includes the below properties:
  - `open`: Toggles the component open/close
  - `helpOpen`: Toggles the help section open/close
  - `helpResizable`: Whether the help section is resizable
  - `escapeKeyAction`: Action to be taken when escape is pressed ('close` is the default action)
- Includes a slot for help section.
- Ability to resize the help section

#### Test Steps

- [ ] `npm run storybook`
- [ ] Navigate to the Fullscreen Takeover story
- [ ] Play around with the properties 

- [ ] `npm run lint` passes.

fullscreen


adamnel:
@bsahitya I'd like to know how this matches up with the separated patterns we ended up with on the design side:
- Focused page, which is an actual full page, and
- Full-screen dialog, which is an overlay on the current page, just... really big
Does this pattern cover both? I'd like to keep naming exact all the way through from design to code

bsahitya:
@adamnel This PR only covers the full-screen dialog. We still need to add the focused page.
I started working on it today. I will probably close this PR and create a new one with all the changes.
 
To keep the naming consistent with design, would it make sense to create a separate component for the focused page?
@owilliams320 what do you think?

[Figma link](https://www.figma.com/design/6w5IaiDUeLVuAPsOBu9JYM/Full-screen-takeover?node-id=4-16&node-type=canvas&t=KXNxCHf399UGBBCE-0) for reference.

PR #2236, opened 2024-09-05
updates Knapsack to `v69.5`

mabry1985:
## Description

Updates Knapsack to `v69.5` and refactors the Angular renderer setup


package.json
```diff
@@ -178,6 +178,7 @@
     "ts-jest": "29.1.1",
     "typescript": "5.3.3",
     "vite": "^4.5.3",
-    "vitest": "^0.27.2"
+    "vitest": "^0.27.2",
+    "webpack": "^5.94.0"
```
mabry1985, line 182: @owilliams320 one thing I forgot to mention in our call was that I needed to explicitly set this webpack dep here otherwise I was getting a conflict between pkgs

apps/knapsack/ks.module.ts
```diff
@@ -0,0 +1,27 @@
+// This module is imported into the ks angular app
+// Use this to add any custom components or modules
+// Note: Providers from the `BrowserModule` have already been loaded.
```
mabry1985, line 3: @owilliams320 note that the `BrowserModule` and `BrowserAnimationModule` are provided by our renderer. 

There was also a similar issue with the `RouterModule` that I ran into, so that has been removed from here as well.

Let us know if you see any issues when implementing your components

apps/knapsack/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
```
mabry1985, line 1: Feel free to adjust this file as needed, the pathing is all we need to have here

apps/knapsack-renderer-angular-app/src/theme-dark.scss
```diff
@@ -1,20 +0,0 @@
-@use '@angular/material' as mat;
```
mabry1985, line 1: We'll need to confirm that we have this file compiled and brought into the asset sets config

owilliams320 (approved):

owilliams320:
Thanks for the updates @mabry1985

PR #2258, opened 2024-10-14
fix(code-editor): add disable scroll option for code editor

bsahitya:
## Description

Add a property to disable scroll in the code editor. The editor height will be dynamically set based on the content when this property is set to true.

### What's included?

- Add `disableScroll` boolean property in code-editor web component

#### Test Steps

- [ ] `npm run storybook`
- [ ] then set disableScroll to true in code-editor
- [ ] Add content to the editor and verify if the editor height is adjusted.

- [x] `npm run lint` passes.

code-editor


owilliams320 (approved):
looks good 👍🏽

PR #2507, opened 2025-05-12
style(theme): update chip token values

owilliams320:
## Description

Updating the chip styling to use the correct tokens

#### Test Steps

- [ ] `npm run start`
- [ ] then this
- [ ] finally this

- [ ] `npm run lint` passes.

##### Before
Screenshot 2025-05-12 at 10 41 10 AM

##### After
Screenshot 2025-05-12 at 10 41 16 AM


adamnel (changes requested):
All good except the checked/selected chip should be `secondary-container` and `on-secondary-container`. Looks like that's coming through for the "secondary" story, but not for the "selectable" story

adamnel (approved):

PR #2525, opened 2025-06-23
Upgrade Covalent to Angular 20.x and Add Density Support to Select Components

owilliams320:
## Description

This PR upgrades Covalent to support **Angular 20.x** and introduces **density support** to key components, enhancing both compatibility and UI flexibility.

### What's included?

- Upgraded Covalent and dependencies to Angular 20.x
- Refactored code to address Angular 20 breaking changes
- Added density support to 
- Ensured backward compatibility and updated styles accordingly

#### Test Steps

- [ ] `npm install` to update dependencies
- [ ] `npm run start` to launch the app
- [ ] Verify component behavior and styling across density modes
- [ ] Run tests and validate build

- [ ] `npm run lint` passes.


baldemarch (dismissed):

libs/components/src/theme/_index.scss
```diff
@@ -44,3 +45,10 @@
       inverse-on-surface
     )};
 }
+
+// --- button density function --- //
+@function density-height($height) {
+  @return calc(
+    #{$height} + (var(--cv-density-mode, 0) * var(--cv-density-interval, 4px))
+  );
+}
```
mikespudwebb, line 54: Is this needed with the denisty.scss function available?

mikespudwebb (approved):
One question. But from what I looked at, seems to make sense. Excited to be able to use the density stuff!

mikespudwebb (approved):

PR #2530, opened 2025-07-01
feat: add CovalentDivider component (DSYS-111)

menhio:
## Description

Added a new CovalentDivider web component that provides flexible visual separation between content. The component supports both horizontal and vertical orientations with multiple size variants and flush options.

### What's included?

- **CovalentDivider component** with TypeScript implementation
- **Multiple size variants**: `full`, `inset`, and `icon` for different spacing needs
- **Orientation support**: Both horizontal and vertical dividers
- **Flush option**: Control whether dividers have default spacing
- **CSS custom properties**: Dynamic styling based on component properties
- **Unit tests**: Basic component functionality tests
- **Component configuration**: Added to component-config.json for build system integration

#### Test Steps

- [ ] Run `npm run storybook` and navigate to "Components/Divider"
- [ ] Test horizontal dividers with different size variants (full, inset, icon)
- [ ] Test vertical dividers with different size variants
- [ ] Toggle flush property to see spacing changes
- [ ] Verify dividers render correctly in both light and dark themes

- [ ] `npm run lint` passes.

Screenshot 2025-06-30 at 11 24 24 p m

### DIFF
--- a/libs/components/src/divider/divider.ts
+++ b/libs/components/src/divider/divider.ts
@@ -1,0 +1,150 @@
+import { LitElement, html } from 'lit';
+
+/**
+ * CovalentDivider
+ *
+ * A divider component that creates a visual separation between content.
+ * Can be used horizontally or vertically with different variants.
+ */
+class CovalentDivider extends LitElement {
+  static properties = {
+    direction: { type: String },
+    size: { type: String },
+    flush: { type: Boolean },
+  };
+
+  constructor() {
+    super();
+    this.direction = 'horizontal';
+    this.size = 'full';
+    this.flush = false;
+  }
+
+  get isVertical() {
+    return this.direction === 'vertical';
+  }
+
+  getSpacing() {
+    const { isVertical, size, flush } = this;
+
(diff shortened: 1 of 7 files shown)


adamnel:
The component seems good, but the previews do not react to the props controls, since they're just showing the whole set of possibilities. That's a great demo but in addition need to break it down so we can play around and see it react

libs/components/src/divider/divider.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * CovalentDivider
+ *
+ * A divider component that creates a visual separation between content.
+ * Can be used horizontally or vertically with different variants.
+ */
+@customElement('cv-divider')
+export class CovalentDivider extends LitElement {
+  static properties = {
```
owilliams320, line 13: i think we will need to follow how lit shows us how to define properties here. 

https://lit.dev/docs/components/properties/

owilliams320 (changes requested):

owilliams320:
Also @menhio looks like the Test and Lint stages are not passing.

PR #2556, opened 2025-09-03
feat(components): adding text style to select

owilliams320:
## Description

Adding text style and variant to support two lines for the select component.

### What's included?

- adding the property `text` that removes default fill style
- adding `twoline` variant that will enable space for a `secondary` slot
- updated select storybook with two line varient and text style updates

#### Test Steps

- [ ] `npm run storybook`
- [ ] then open the two line select story
- [ ] finally open the properties for the story and select "text" to see the new text style

- [ ] `npm run lint` passes.

Screenshot 2025-09-02 at 4 45 52 PM
Screenshot 2025-09-02 at 4 44 37 PM

### DIFF
--- a/libs/components/src/app-shell/app-shell.stories.js
+++ b/libs/components/src/app-shell/app-shell.stories.js
@@ -1,6 +1,5 @@
 import tableRowSelectionContent from '../../stories/demos/table-row-selection.content.html?raw';
 import '../data-table/data-table.stories.scss';
-
 import './app-shell';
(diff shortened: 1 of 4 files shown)


adamnel:
@owilliams320 Looks great - for our purposes we want a style that looks similar to the "filled" one but without the underline and with rounded corners on the bottom as well. How would you handle that, here permanently or as an override each place we use it? I don't really forsee us ever using this style as is

PR #2572, opened 2025-09-18
Fix/cv textfield loading

bsahitya:
## Description

- Add support for a loading state in `cv-textfield`
- Support `updateOn` form control strategy in the `CovalentTextfieldValueAccessorDirective`

### What's included?

- Validators are applied based on the `updateOn` strategy provided in the form control
- `cv-textfield` supports a loading state with the below properties:
  - `loading`: boolean (sets the loading state)
  - `loaderDensity`: number (sets the size of the spinner)

#### Test Steps

- [ ] `npm run storybook`
- [ ] then view Textfield -> Loading story
- [ ] finally verify the loading state

- [x] `npm run lint` passes.

covalent-form

textfield

### DIFF
--- a/libs/angular/common/src/directives/web-components/form-control.directive.spec.ts
+++ b/libs/angular/common/src/directives/web-components/form-control.directive.spec.ts
@@ -68,7 +68,7 @@ describe('CovalentTextfieldValueAccessorDirective', () => {
   it('should update form control when textfield value changes', () => {
     const textfield = fixture.nativeElement.querySelector('cv-textfield');
     textfield.value = 'new value';
-    textfield.dispatchEvent(new Event('change'));
+    textfield.dispatchEvent(new Event('input'));
     fixture.detectChanges();
     expect(component.control.value).toBe('new value');
   });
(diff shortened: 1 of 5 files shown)


adamnel:
The space taken up when this shows up should be subtracted from the field width, so that that overall width does not change.

How does this interact with fields that have a trailingIcon?

bsahitya:
> The space taken up when this shows up should be subtracted from the field width, so that that overall width does not change.
> 
> How does this interact with fields that have a trailingIcon?

👍🏽  I'll update the width.

Currently, when both loading and trailingIcon are set, only the loading state is displayed.
Would you prefer we show both simultaneously?

adamnel:
> Currently, when both loading and trailingIcon are set, only the loading state is displayed. Would you prefer we show both simultaneously?

Yeah that's fine to just replace the existing icon with the loader

bsahitya:
> > Currently, when both loading and trailingIcon are set, only the loading state is displayed. Would you prefer we show both simultaneously?
> 
> Yeah that's fine to just replace the existing icon with the loader

@adamnel I made the changes to the field width. Please take a look and let me know your thoughts:
https://630f72168004b8037704328f-vqtgkezbng.chromatic.com/?path=/story/components-text-field--loading

